Issue #1, opened 2020-06-08
Principles & Goals of Project


JordanMartinez:
## Principles

- each recipe is a stand-alone Spago project that solves only 1 problem
- no or limited pedagogical explanations. If anything requires more explanation, link to other resources
- in the recipe's ReadMe
    - indicate which libraries are used/explained/demonstrated
    - **Added 6/7/2020**: Describe the problem the recipe solves

JordanMartinez:
Remaining questions to be answered:
- How does Rust's cookbook work/look? (Copy success)
- If someone submits a recipe, what should the folder that contains the recipe be called, so that it doesn't lead to name clashing / merge conflicts?
- If multiple recipes on the same/similar topics are submitted, do we merge them into one? Keep them separate?

JordanMartinez:
As I thought about this more today, I realized there's a tradeoff here in how we could approach things. In other words, whether to use 1 Spago project for each recipe or 1 Spago project for the entire repo:

## Use 1 Spago project for the entire repo

Pros:
- simpler CI build times and maintenance. Install `purescript`, `spago`, `parcel`, etc. and just run `spago test`

Cons:
- actual libraries used by example aren't shown
- merge conflicts might occur (A wants to use package set X whereas B wants to use package set Y)

## Use 1 Spago project for each recipe

Pros:
- each is stand-alone, so any modifications to it won't cause merge conflicts and whatnot
- libraries used are clearly visible
- different package set versions could be used in case one recipe becomes slightly outdated

Cons:
- We will compile the same code (e.g. Prelude) many times, slowing down CI. If we get too many recipes, we'll hit the CI build timeout. 

The Con to this one is not an issue right now, but could be in the future. Perhaps a possible solution is to run multiple builds that each test only a subset of the entire repo's recipes...?

milesfrain:
I think we should do the single spago project strategy in the interest of keeping CI fast.
Including all libraries for the latest package set seems fine:
* It's straightforward to figure out what libraries are required for a given import (and if there's any ambiguity, that can be noted in that example's readme).
* We should aim to have everything work with the latest package set.
* TryPursuit uses this system.

It would be super cool if PS could maintain a global cache of versioned prebuilt libraries that are automatically linked to from each independent project. This is what Julia does, and it's pretty sweet.

thought2:
I like the idea of having one `spago.dhall` file per recipe.

[...]

PR #4, opened 2020-06-10
Setup Travis CI and see if it works

JordanMartinez:


JordanMartinez:
A few questions I have with this implementation:
- How do I get CI to fail if even one recipe's `ci` NPM script returns a non-zero exit?
- Should I install purescript/spago/parcel globally? Or should each recipe do that? I prefer the former because I want to reduce the length of time to test one recipe
- Any idea for how we can reduce the amount of time spent recompiling all the same code (e.g. by sharing the output folder)?

milesfrain:
Let's do single project as mentioned in #1 

I think just ensuring basic `spago build` works is enough for testing (no need for production build).

Also, running just a single script or command in travis would be ideal, since then it's easy to run the same set of tests locally.

I'd be happy to dive into this more tomorrow.

For reference, `set -e` in scripts is one way catch an error in any of the commands in the script.

JordanMartinez:
Closing in favor of #6

PR #16, opened 2020-06-11
Remove need to use `cookbook:` prefix when specifying recipe

### LINKED ISSUE #13: Remove 'cookbook:' prefix when specifying recipe
**Description of Problem**

A shorter command is easier to type/modify/read later than a longer one. 

We currently have this command:
```
npm run build-dev --cookbook:recipe=RoutingHashLog
```
I would prefer to use this command and omit the "cookbook:" prefix:
```
npm run build-dev --recipe=RoutingHashLog
```

I believe this can be done by removing the 'name' field in the package.json file.

JordanMartinez:
Fixes #13

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,4 @@
 {
-  "name": "cookbook",
   "license": "MIT",
   "devDependencies": {
     "parcel": "^1.12.4",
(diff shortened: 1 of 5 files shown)


milesfrain:
Nice. Good job checking into this. Not sure what I tried earlier that wasn't working, because I really didn't want to to add the 'name' field.

PR #18, opened 2020-06-11
Change `meta issue` to `cookbook issue`

milesfrain:
Also figured it would be nice to make this more freeform rather than duplicate the recipe sections


JordanMartinez:
LGTM! I'll rename the `meta` label to `cookbook`

PR #19, opened 2020-06-12
Migrate from GitHub CI to Travis CI

### LINKED ISSUE #9: Determine which CI provider to use
**Description of Problem**

See [this comment in #6](https://github.com/JordanMartinez/purescript-cookbook/pull/6#discussion_r438515055) for more context. We're currently using GitHub CI. I'm not sure I want to continue using that because of being less familiar with that. However, I also don't want to use Travis because "everyone does that."

I'd like to discuss this further as to which one we should use and why.

JordanMartinez:
Fixes #9


JordanMartinez:
@milesfrain I'll leave this to you to merge. Do so if you're fine with it.

PR #20, opened 2020-06-12
Add note about not adding/overriding packages in `packages.dhall` file.

### LINKED ISSUE #8: Only use packages in latest package set
In other words, one cannot add or override packages in `packages.dhall`. See [this comment](https://github.com/JordanMartinez/purescript-cookbook/pull/6#discussion_r438516004) for more context.

JordanMartinez:
Fixes #8 

If this "contributing a new recipe" section gets too big or should appear in a CONTRIBUTING.md file, we can do that later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,13 +6,21 @@ The recipes folder contains all the recipes in this repo in no particular order.
 
 ## Contributing new recipes
 
+### Principles
+
+#### Only use packages in the latest package set release
+
+In other words, we do not allow anyone to override or add packages in the `packages.dhall` file. If one overrided a given package to make Recipe A work, it might break Recipe B that also relies upon that package. If one added a package that isn't later maintained, it might prevent our CI from passing, which can delay new recipes from being added to this cookbook.
+Thus, if you want to use a package that's not in the package set, please add it to the package set. If you want to override a package, please update it in the package set. No exceptions.
+
+### Instructions
+
 Follow these instructions for contributing new recipes:
 1. Open a new issue making a request for a recipe. This helps avoid two developers working on the same recipe independently.
 1. Pick an existing recipe to duplicate as a starting point. The `Template` recipe is probably the simplest".
 1. Rename the copied folder to a short description of the problem you're trying to solve using PascalCase (e.g. RoutingPushHalogen)
 1. If your recipe is incompatible with the browser enviornment (e.g. reading local files with node), delete the `dev` directory. Note that just logging to the console **is** supported in the browser.
 1. Use your recipe name to prefix all modules. Do a quick search to find all instances of the old recipe name to replace (e.g. `grep -r RoutingPushHalogen .`).
-1. If needed, update the `packages.dhall` file to include more packages.
 1. Install needed dependencies via `spago install <packageName>`
 1. Implement your recipe
 1. Rember to update your recipe's readme
--- a/README.md
+++ b/README.md
@@ -10,8 +10,14 @@ The recipes folder contains all the recipes in this repo in no particular order.
 
 #### Only use packages in the latest package set release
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,13 +6,21 @@ The recipes folder contains all the recipes in this repo in no particular order.
 
 ## Contributing new recipes
 
+### Principles
+
+#### Only use packages in the latest package set release
+
+In other words, we do not allow anyone to override or add packages in the `packages.dhall` file. If one overrided a given package to make Recipe A work, it might break Recipe B that also relies upon that package. If one added a package that isn't later maintained, it might prevent our CI from passing, which can delay new recipes from being added to this cookbook.
```
milesfrain, line 13: Thinking this might read better as bullets:
> * Overriding a package to make Recipe A work might break Recipe B ...
> * Packages outside of the package set are more likely to be neglected ...

The first situation would be caught by CI though.

A more realistic scenario is Recipe A making it through CI with an overridden package, which then _prevents_ a new Recipe B from passing CI with the default package set. Maybe that's the situation you have in mind.

Also wondering about overrides to more recent versions of packages already tracked in the package set. For example can new Halogen Hooks recipes get merged while we're waiting for the latest version to be picked up? I guess it's fair to just those recipe PRs sit until the corresponding package set PR is merged.

JordanMartinez: > Also wondering about overrides to more recent versions of packages already tracked in the package set. For example can new Halogen Hooks recipes get merged while we're waiting for the latest version to be picked up? I guess it's fair to just those recipe PRs sit until the corresponding package set PR is merged.

Yeah, we'd wait to merge it until later. 

> Thinking this might read better as bullets:

I agree.

> The first situation would be caught by CI though.
> 
> A more realistic scenario is Recipe A making it through CI with an overridden package, which then prevents a new Recipe B from passing CI with the default package set. Maybe that's the situation you have in mind.

That's true. To clarify more points, I think there's a maintenance burden we don't pay for if we only support recipes which can run using the regular package set. The other concern is if someone adds/overrides a malicious package. While we don't have a strong vetting process, it at least goes through more eyes this way.

JordanMartinez: I've accounted for your feedback. Please let me know your thoughts.

README.md
```diff
@@ -6,13 +6,27 @@ The recipes folder contains all the recipes in this repo in no particular order.
...
+#### Only use packages in the latest package set release
+
+In other words, we do not allow anyone to override or add packages in the `packages.dhall` file for the following reasons:
+- Reduces the burden of maintenance for maintainers of this repo
+- Overriding package A to make Recipe X work now might prevent Recipe Y from working (which doesn't need the override)
+- Somewhat reduces the attack surface of malicious recipes (e.g. including a malicious package via addition/override)
+
+Thus, if you want to use a package that's not in the package set, please add it to the package set. Any PRs that use packages not in the package set will not be merged until all of their packages are in the package set.
+
+By implication, recipes that use packages that get dropped from the package set will be moved to a temporary holding folder. If the recipe doesn't get updated within a month, the recipe will be dropped.
```
JordanMartinez, line 20: @milesfrain Does this "recipe-dropping" policy seem reasonable to you? I think in times of breaking changes, we would extend this to be longer (e.g. 3 months?)

milesfrain: Sure. Do we need a concrete timeline for dropping? Thinking they could just hang out in the broken folder until someone decides to fix them or overwrite the recipe to use an alternative package.

JordanMartinez: Yeah, I think moving them to a 'broken recipes' folder makes more sense. People on later versions can still find them helpful / people trying to update them have less work to do.

Could you merge both this PR and the next one? I got work starting.

PR #21, opened 2020-06-12
Try simpler travis

milesfrain:
Wanted to see if this works too on main repo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,29 +1,7 @@
 dist: bionic
-language: minimal
-  # Use `nvm` to install latest version of NodeJS
-
-# Unless we specify "sudo: false", build runs in VM, not container
-
-# Limit the number of commits it will clone
-git:
-  depth: 3
-
-# Only do CI for the default branch
-branches:
-  only:
-  - master
-
-before_install:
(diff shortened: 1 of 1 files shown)


milesfrain:
Follow-up to #9 and #19 

Simpler seems better.

Also I recall reading something about explicit caching in travis not being recommended, since it's either enabled by default for the things that should be cached, and lots of times manually caching the wrong stuff hurts performance https://docs.travis-ci.com/user/caching/

JordanMartinez:
I'd prefer simpler, so this looks good.

Clarification for future readers: [`node_modules` is cached by default by Travis](https://docs.travis-ci.com/user/caching/#npm-cache) and [restoring the cache requires download the cache, which may not be faster than just downloading those things yourself](https://docs.travis-ci.com/user/caching/#things-not-to-cache). So, these config values are not needed:
```yml
cache:
  directories:
    - .spago
    - node_modules
```

However, should we still keep this config?
```yml
# Only do CI for the default branch
branches:
  only:
  - master
```

JordanMartinez:
Actually, let's address that in a separate PR. I'd like to keep iterating.

Issue #24, opened 2020-06-12
Reconsider the tools used to run recipes to account for multiple parties' interests


JordanMartinez:
While trying to research multi-line npm scripts, I found this package: [script-launcher](https://www.npmjs.com/package/script-launcher). Thoughts on using it here?

**Edit:** See the next comment I make for the heart of this matter.

milesfrain:
Yes, that looks really nice. Seems like we could also specify web vs node compatibility per recipe within the menu config.

JordanMartinez:
Ok, let me put this a different way, starting from the top-down principle-first approach.

- [Users] - We want users to be able to clone this repo and be able to easily run whichever recipe they desire. They should not have to worry about installing npm dependencies unless a recipe requires it (e.g. reduce possible attack surface). We want to eliminate annoying errors when possible (i.e. #10) or minimize them by providing helpful action-oriented error messages. The commands to run the scripts should read from 'left-to-right,' the order in which our brains will think of those recipes. For example, something like `npm run MyRecipe --target=browser` or `npm run MyRecipe --target=production-thingy-majiggy`. Not right-to-left as we currently have things npm run production-thingy-majiggy --recipe=MyRecipe`. While these could be standardized (e.g. #12), it wouldn't be required in this approach.
- [Reciper-Makers] - We want those who write recipes to have full freedom in how they implement their recipe. If they want to have 10 different ways to build a recipe, that should be possible. For example, a complicated recipe might have 2+ targets (i.e. build the webapp server, build the client part, build the file server, build the database server). While I would likely reject a recipe that does all of that (rather than multiple recipes that do that individually), it should still be possible and easy to do. Moreover, the scripts themselves should be stand-alone and be stored within the same folder as their recipe.
- [Maintainers] - We want maintainers to be able to add/tweak/modify one recipe without affecting all others. We want CI to be simple and clear. In other words, let's say I had a recipe in the `MyCoolRecipe` folder that suddenly became broken at some point (e.g. package set issue, breaking release in a library, breaking release in PS, etc.). The only thing I should have to do is move the `MyCoolRecipe` folder from `src/` into `broken/`. CI should pass after that if nothing else is affected.

Our options are currently
- `npm` scripts. These suck because they don't scale well
- `script-launcher`. This provides a better UI, but is still based on `npm` scripts. I'm not sure how well it would scale, and it needs more research
[...]

PR #27, opened 2020-06-13
Add makefile that uses pattern matching strategy

milesfrain:
**Edit by Jordan:** Partially addresses #24 

Here's one makefile approach.

There are some options for command format. Currently using the first one:
* `make Template-run`
* `make runTemplate`
* etc...

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,0 +1,95 @@
+# ------------- Common --------------
+
+# Favor local npm devDependencies if they are installed
+export PATH := node_modules/.bin:$(PATH)
+
+# Use `PHONY` because target name is not an actual file
+.PHONY: build info
+
+# Builds everything. No recipe required.
+# A build is necessary before parcel commands.
+# Users could just run `spago build` instead of `make build`,
+# unless they want local npm version and don't want to run `npx spago build`.
+build:
+	spago build
+
+# Prints version and path information.
+# For troubleshooting version mismatches.
+info:
+	which purs
+	purs --version
+	which spago
+	spago version
+	which parcel
+	parcel --version
+
+# Tests if recipe actually exists.
+# This should be a dependency of all entry targets
+recipes/%:
+	test -d $* || { echo "Recipe $* does not exist"; exit 1;}
+
+# -------- Pattern matching strategy -----------
+
+# Usage:
+#
+# make Template-run
+# make Template-serve
+# make Template-buildDev
+# make Template-buildProd
+
+# Targets for all recipe build operations
+recipes := $(shell ls recipes)
+recipesRun := $(foreach r,$(recipes),$(r)-run)
+recipesServe := $(foreach r,$(recipes),$(r)-serve)
+recipesBuildDev := $(foreach r,$(recipes),$(r)-buildDev)
+recipesBuildProd := $(foreach r,$(recipes),$(r)-buildProd)
+
+# Use `PHONY` because target name is not an actual file
+.PHONY: recipesRun recipesServe recipesBuildDev recipesBuildProd buildAllDev buildAllProd
+
+# Helper functions for generating paths
+main = $1.Main
+recipeDir = recipes/$1
+
+devDir = $(call recipeDir,$1)/dev
+devHtml = $(call devDir,$1)/index.html
+devDistDir = $(call recipeDir,$1)/dev-dist
+
+prodDir = $(call recipeDir,$1)/prod
+prodHtml = $(call prodDir,$1)/index.html
+prodJs = $(call prodDir,$1)/index.js
+prodDistDir = $(call recipeDir,$1)/prod-dist
+
+# Runs recipe as node.js console app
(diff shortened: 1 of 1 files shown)


milesfrain:
Here's another option which is a bit simper, but has the following dowsides:
* Longer command due to passing `recipe` arg: `make recipe=Template run`
* Recursively calls `make` in a sloppy (non-idiomatic way) for `buildAllDev` / `buildAllProd`.

```make
# ------------- Common --------------

# Favor local npm devDependencies if they are installed
export PATH := node_modules/.bin:$(PATH)

# Use `PHONY` because target name is not an actual file
.PHONY: build info

# Builds everything. No recipe required.
# A build is necessary before parcel commands.
# Users could just run `spago build` instead of `make build`,
# unless they want local npm version and don't want to run `npx spago build`.
build:
	spago build

# Prints version and path information.
# For troubleshooting version mismatches.
info:
	which purs
	purs --version
	which spago
	spago version
	which parcel
	parcel --version

# Tests if recipe actually exists.
# This should be a dependency of all entry targets
recipes/%:
	test -d $* || { echo "Recipe $* does not exist"; exit 1;}

# ---------- `recpipe` strategy ----------------
#
# Example usage:
#
# make recipe=Template run
# make recipe=Template serve
# make recipe=Template buildDev
# make recipe=Template buildProd

# Helper-variables derived from `recipe` arg
main := $(recipe).Main
recipeDir := recipes/$(recipe)

devDir := $(recipeDir)/dev
devHtml := $(devDir)/index.html
devDistDir := $(recipeDir)/dev-dist

prodDir := $(recipeDir)/prod
prodHtml := $(prodDir)/index.html
prodJs := $(prodDir)/index.js
prodDistDir := $(recipeDir)/prod-dist

# Use `PHONY` because target name is not an actual file
.PHONY: run serve buildDev buildProd buildAllDev

# Runs recipe as node.js console app
run: $(recipeDir)
	spago run --main $(main)

# Launches recipe in browser
serve: $(recipeDir) build
	parcel $(devHtml) --out-dir $(devDistDir) --open

# Uses parcel to quickly create an unminified build.
# For CI purposes.
buildDev: export NODE_ENV=development
buildDev: $(recipeDir) build
	parcel build $(devHtml) --out-dir $(devDistDir) --no-minify --no-source-maps

$(prodDir):
	mkdir -p $(prodDir)

$(prodHtml): $(prodDir)
	cp $(devHtml) $(prodDir)

# Creates a minified production build.
# For reference.
buildProd: $(recipeDir) $(prodHtml)
	spago bundle-app --main $(main) --to $(prodJs)
	parcel build $(prodHtml) --out-dir $(prodDistDir)

# Creates all dev builds - for CI
buildAllDev:
	for d in recipes/*; do \
	    if [ -d $$d/dev ]; then \
	    	make recipe=$$(basename $$d) buildDev; \
	    fi \
	done

# Creates all prod builds - for CI
buildAllProd:
	for d in recipes/*; do \
	    if [ -d $$d/dev ]; then \
	    	make recipe=$$(basename $$d) buildProd; \
	    fi \
	done
```

JordanMartinez:
First, thanks for figuring this out! I haven't used `make` before and I'm not very strong in `bash`.

> make Template-run

I like this approach to things. Less to type and it's pretty readable.

JordanMartinez:
I think running just `make` should print the list of other scripts one could run (e.g. `make build` `make Template-*`, etc.). Then one could easily see all recipes that one could build as well as other things (e.g. clean, ci, help, etc.)

milesfrain:
Sure. That addition should be pretty straightforward. Should we make that a separate task?

JordanMartinez:
Seems like GNU make doesn't include a `list-scripts` task. I found [this SO answer](https://stackoverflow.com/a/26339924), but it doesn't work with our patterns approach. Here's the current output of `make` when `list` is the first rule using this PR:
```
$ make
build
buildAllDev
buildAllProd
info
recipesBuildDev
recipesBuildProd
recipes/RoutingHashHalogen
recipesRun
recipesServe
RoutingHashLog
RoutingPushHalogen
Template/prod
```

milesfrain:
```makefile
.PHONY: list      
list:                 
  @echo === RECIPES ===
  @echo $(foreach r,$(recipes),$(r)-{run,serve}) | tr ' ' '\n'
```

Produces:
```
make list
=== RECIPES ===
RoutingHashHalogen-{run,serve}
RoutingHashLog-{run,serve}
RoutingPushHalogen-{run,serve}
Template-{run,serve}
```

Some additional logic can be added to determine whether `run` or `serve` are valid commands.

JordanMartinez:
Well, let's add that version of the recipe list in, and then get this merged. I believe this will cause problems later, but let's take that risk for now, so we can continue iterating on other parts of this repo.

Since this is a simple fix, I'll merge this PR and then add that list myself afterwards.

PR #28, opened 2020-06-14
rename Template to HelloWorld

milesfrain:


JordanMartinez:
I'm "iffy" on this. Having it names as "template" makes it pretty clear why it's there: it's a starter to use for building recipes. Changing it to HelloWorld makes this less obvious.

However, you pointed out earlier that it may be easier for someone to copy a different recipe, make a few modifications, and then push that recipe. So, the idea of a "template" recipe might not be as useful as I thought.

On another hand, "hello world" usually indicates the simplest way to do X. So, I guess we could refer to it as the "hello world" of writing recipes.

JordanMartinez:
Let's present this as both a "hello world" for PureScript but also a "hello world" for recipe-writing.

PR #30, opened 2020-06-14
Feature Instructions before Principles

milesfrain:
Instructions should probably come first.

We can point folks to "Principles" (maybe rename this to "Additional Guidelines") if there are PR issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,19 @@ The recipes folder contains all the recipes in this repo in no particular order.
 
 ## Contributing new recipes
 
+### Instructions
+
+Follow these instructions for contributing new recipes:
+1. Open a new issue making a request for a recipe. This helps avoid two developers working on the same recipe independently.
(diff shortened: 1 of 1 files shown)


JordanMartinez:
I'm guessing you are proposing this because the reader will find the instructions for submitting a recipe sooner when they are reading through the Readme. The problem with doing this is that the reader might believe these instructions are illustrative, not exhaustive and strict (i.e. they might modify `packages.dhall`). 

So, if these two changes were made, I would be find with this:
1. Include a note that "rationale" for these instructions are described in the 'principles' section (with a link to it)
2. A note after the `spago install <packageName>` command that `packages.dhall` must remain unchanged (aside from upgrading to the next package set).

JordanMartinez:
> We can point folks to "Principles" (maybe rename this to "Additional Guidelines") if there are PR issues.

I think all of this should go into a Contributing.md file. I'll create a separate PR. Then the ReadMe will only display recipes and instructions to read the contributing.md file if they want to add a recipe.

PR #33, opened 2020-06-14
Edit packages in table

milesfrain:
`prelude` is implied, and doesn't need to be explicitly listed in `spago.dhall`.

I'm wondering if `console` and `effect` are worth listing either. It makes this list pretty busy. Maybe we should just list additions beyond what's available by default after `spago init`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ If you want to contribute a new recipe, see the [CONTRIBUTING.md](./CONTRIBUTING
 
 | Recipe | Description | Packages | Node | Browser |
 | - | - | - | - | - |
-| [HelloWorld](./recipes/HelloWorld) | Prints "Hello World" to console | `effect`, `console`, `prelude` | x | x |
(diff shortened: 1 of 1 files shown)


JordanMartinez:
Part of me wonders whether libraries should be included in this table at all. What if the table just had the following columns: Name, Description, Runs on Node, Runs on Browser?

For the time being, I'll merge this in because this is the current approach we're taking.

milesfrain:
Listing libraries is nice for letting users find all `routing` or `halogen` recipes, although they could also just grep for imports.

Actually, it would probably be best to us to just automatically build an index of all imported packages. (stretch goal).

So I'm fine with deleting the "Packages" column.

JordanMartinez:
> Actually, it would probably be best to us to just automatically build an index of all imported packages. (stretch goal).

I think this is actually possible since `purs graph` will output the following:
```
{"Path.To.Module":{path:".spago/package/src/Path/To/Module.purs", depends: []}}
```
We could extract the packages it depends upon based on the modules it uses.

PR #35, opened 2020-06-15
Use a separate spago.dhall per recipe

milesfrain:
Implementing suggestion in https://github.com/JordanMartinez/purescript-cookbook/issues/1#issuecomment-644157098

Also nicely addresses some points in #33 

There's a minor annoyance where the recipe name must be duplicated with `sources` in `spago.dhall`. See https://github.com/purescript/spago/issues/648 for more details.

@JordanMartinez Would you like to take a stab at the corresponding `makefile` changes for this? Let me know if you want hints.


JordanMartinez:
I probably should to get more familiar with `make`. However, I won't be able to get to that until tomorrow at the very earliest.

JordanMartinez:
Also, is it possible for you to do your work on a branch on this repo rather than on your repo? Then, I can push new commits to this branch without having to submit a PR to your repo. That should speed up some of our development.

JordanMartinez:
See my [PR](https://github.com/milesfrain/purescript-cookbook/pull/3) to your repo which updates the `makefile`

JordanMartinez:
Merged!

> The individual recipe readmes are now out of date. Thinking it would be good to minimize those and strip out "Quick Start" and "Packages Used". Perhaps just leave a short summary of expected behavior.

Yeah, we'll need to update that. I'll open an issue to track that.

milesfrain:
@JordanMartinez

FYI, This syntax from https://github.com/JordanMartinez/purescript-cookbook/pull/35/commits/fb37b675e0bc16a38603285ba9d71f75f4d6345b#diff-cd5dbf3629c558b0104ba8f0937d6816R76 is not quite right.
```make
$(call $*-build,%)
```

It should just be replaced with:
```make
%-build
```

Otherwise it attempts to call a function with the same name as this prereq and then gets back an empty string.

Included a fix for this in #58

JordanMartinez:
Gotcha. Thanks!

PR #46, opened 2020-06-18
Add script to automatically generate table of recipes

### LINKED ISSUE #34: Autogenerate Readme's list of recipes
See [this comment](https://github.com/JordanMartinez/purescript-cookbook/pull/31#issuecomment-643796213) for more context.

milesfrain:
Fixes #34

### DIFF
--- a/RecipeTable.md
+++ b/RecipeTable.md
@@ -1,8 +1,0 @@
-# Recipes
-
-| Recipe | Description | Node | Browser |
-| - | - | - | - |
-| [HelloWorld](recipes/HelloWorld) | This recipe shows how to run a simple "Hello world!" program in either the node.js or web browser console. | :heavy_check_mark: | :heavy_check_mark: |
-| [RoutingHashHalogen](recipes/RoutingHashHalogen) | This recipe shows how to use `purescript-routing` to do client-side hash-based routing in a Halogen-based single-page application (SPA). |   | :heavy_check_mark: |
-| [RoutingHashLog](recipes/RoutingHashLog) | This recipe demonstrates hash-based routing with `purescript-routing`. No web framework is used. | :heavy_check_mark: | :heavy_check_mark: |
-| [RoutingPushHalogen](recipes/RoutingPushHalogen) | This recipe shows how to use `purescript-routing` to do client-side push-state routing in a Halogen-based single-page application (SPA). |   | :heavy_check_mark: |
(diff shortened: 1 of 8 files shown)


milesfrain:
Some follow-up CI tasks are to either:
1. Run `scripts/generateRecipeTable.sh > referenceTable.md` as first step of CI and fail if `referenceTable.md` is different than `RecipesTable.md`. This means that the contributor failed to update the recipe table themselves.
2. Automatically generate and overwrite this table as part of CI. I don't like this as much because the PR won't be an accurate preview of what actually gets comited to the repo. One way this could be an issue is if the line that the recipe description lives on changes. Then we'd end up with an incorrect description.

So planning on the first option as a next step.

RecipeTable.md
```diff
@@ -0,0 +1,8 @@
+# Recipes
```
milesfrain, line 1: Probably don't need this heading. Considering removing it.

scripts/generateRecipeTable.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+echo "# Recipes"
+echo
+echo "| Recipe | Description | Node | Browser |"
+echo "| - | - | - | - |"
+for d in recipes/*; do
+  base=$(basename $d)
+  description=$(sed '3q;d' $d/README.md)
```
JordanMartinez, line 9: We'll need to add instructions that clarify that the Recipe description must always be on the 3rd line in the file and cannot span across multiple lines.

JordanMartinez:
I'd like to request one minor addition and one major change to this approach. 

### Minor Change

Rather than asking the person to run that script, why not add a `make` command that does that? Then we can tell them to run `make readme`, which sounds like what it does.

### Major Change

I'd prefer if the Recipes ToC is on the same page as this repo's Readme. So, what if we reconstructed this repo's readme every time we ran the above script? Here's how it could be done:
1. Store the first unchanging part of the Readme in one file
1. Read that file and overwrite this repo's `Readme.md` with that content
1. Run the script you have above and append the result to the `Readme.md` file

JordanMartinez:
I've also reordered the columns, so that the backend info appears first. [Here's the rich diff to the file](https://github.com/JordanMartinez/purescript-cookbook/pull/46/files?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8)

milesfrain:
I think this is everything on the checklist. The CI check should be a separate PR

JordanMartinez:
Ok. I'll merge then.

PR #50, opened 2020-06-18
Check if recipe table was properly updated

### LINKED ISSUE #49: Detect recipe table mismatch
https://github.com/JordanMartinez/purescript-cookbook/pull/46#issuecomment-646231413

milesfrain:
Resolves #49

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ node_js: 12
 install:
   - npm install
 script:
+  - diff README.md <(scripts/generateRecipeTable.sh)
   - make buildAll
   - make buildAllDev
 cache:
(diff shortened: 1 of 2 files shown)


milesfrain:
Demonstrated working in [this job](https://travis-ci.com/github/JordanMartinez/purescript-cookbook/builds/172118382)

It was a little surprising to see the script continue after this failure. It still returns the correct failure status though.
We can get fail-fast behavior by adding `set -e` on the first line (https://github.com/travis-ci/travis-ci/issues/1066)
Although upon further review, I think it would be nice to let the build continue so contributors can see any other errors they might have in that same PR.

JordanMartinez:
> Although upon further review, I think it would be nice to let the build continue so contributors can see any other errors they might have in that same PR.

I see pros and cons to this... I think it's better to leave this on and only turn it off when we hit CI issues. Thoughts? (per #52)

PR #52, opened 2020-06-19
Add shell and make flags

### LINKED ISSUE #51: Add additional shell and make flags to improve reliability of `makefile`
Add these flags based on [Your makefiles are wrong](https://tech.davis-hansson.com/p/make/) and the overview of [the shell flags](http://redsymbol.net/articles/unofficial-bash-strict-mode/)
```make
.SHELLFLAGS := -eu -o pipefail
MAKEFLAGS += --warn-undefined-variables
MAKEFLAGS += --no-builtin-rules
```

While "Your makefiles are wrong" also recommends adding the `-c` shell flag, [which might refer to the `-C` flag](https://ss64.com/bash/set.html) flag, it seems like that isn't desired because we overwrite the repo's README.md file.

JordanMartinez:
Fixes #51

### DIFF
--- a/makefile
+++ b/makefile
@@ -19,6 +19,13 @@ endif
 #    the pipeline will emit an exit code of 0.
 .SHELLFLAGS := -eu -o pipefail
 
+# Emits a warning if you are referring to Make variables that don’t exist.
+MAKEFLAGS += --warn-undefined-variables
+
+# Removes a large number of built-in rules. Remove "magic" and only do
+#    what we tell Make to do.
+MAKEFLAGS += --no-builtin-rules
+
 # Use `PHONY` because target name is not an actual file
 .PHONY: list readme info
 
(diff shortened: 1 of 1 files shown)


JordanMartinez:
Looks like this PR [broke the build](https://travis-ci.com/github/JordanMartinez/purescript-cookbook/builds/172142120#L229)...

IIRC, the default shell for `make` is `sh`, not `bash`. This can be fixed by setting the default shell to bash.

milesfrain:
> Looks like this PR [broke the build](https://travis-ci.com/github/JordanMartinez/purescript-cookbook/builds/172142120#L229)...

Oof. That's pretty bad to have a failure still show as passing in CI. I guess `make` is returning `0` (passing) with that error?

JordanMartinez:
Something like that. I've already fixed this in master.

PR #54, opened 2020-06-19
Set default shell to bash; add -c to shell flags

### LINKED ISSUE #51: Add additional shell and make flags to improve reliability of `makefile`

JordanMartinez:
Really fixes #51 by fixing problems introduced by #52.

### DIFF
--- a/makefile
+++ b/makefile
@@ -21,7 +21,7 @@ SHELL := bash
 #    pipeline (e.g. `cat file.txt | grep 'foo'`) will be used as the exit
 #    code for the entire pipeline. If all exit codes of a pipeline are zero,
 #    the pipeline will emit an exit code of 0.
-.SHELLFLAGS := -eu -o pipefail
+.SHELLFLAGS := -eu -o pipefail -c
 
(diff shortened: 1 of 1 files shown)


milesfrain:
I'm not sure what the `-c` flag does for us in this case:

https://linux.die.net/man/1/bash

I assume it's this:
> -c string
If the -c option is present, then commands are read from string. If there are arguments after the string, they are assigned to the positional parameters, starting with $0.

PR #55, opened 2020-06-19
Remove some old makefile comments

milesfrain:


milesfrain:
Hmm. I notice a seemingly-unnecessary `Branch` job created. Maybe restricting travis to `master` branch only is the right thing to do.

image

JordanMartinez:
Yeah... I'll open an issue to investigate that further.

Issue #56, opened 2020-06-19
Recipes with dependencies beyond spago


Kamirus:
Hi, many thanks for creating such a project!

We thought about adding some recipes about [purescript-selda](https://github.com/Kamirus/purescript-selda), but it would require more dependencies like npm dependencies and (ideally) PostgreSQL.

Is it a place for such recipes? Do you plan to support it?

JordanMartinez:
The `npm` dependencies shouldn't be an issue. However, were you thinking of setting up an entire database in this repo? If so, that could get tricky...

JordanMartinez:
Regarding `npm` dependencies, there are two ways we could do this:
1. Stores all dependencies in the repo's `package.json` file. This is the easiest to do because all other code should remain the same. However, it doesn't help others know which `npm` dependencies are installed for a given recipe. Moreover, if one recipe should rely on version A and another rely on version B of the same package, it will cause problems.
2. Store a recipe's dependencies in the recipe folder. This is better for the end-user but makes maintenance a bit harder to do. If someone `require`s a dependency in their FFI, will it resolve to the correct module?

JordanMartinez:
@milesfrain I think this is the last issue that needs to be figured out. Once done, we could clean things up one more time and then I think this repo will be ready for recipe requests. While #63 might be a problem, I'm less worried about it.

milesfrain:
My plan is to cross this bridge when incorporating a recipe that requires these npm dependencies. We could give that a shot now with some `affjax` + `xhr2` recipes.

JordanMartinez:
I've tried to add a recipe for [`purescript-bigints`](https://pursuit.purescript.org/packages/purescript-bigints/) and have encountered a number of issues.

Here's the outcomes I'm trying to reach
1. each recipe folder has its own `package.json` if it needs to install npm dependencies. This tells users which npm dependencies and versions are being used in that recipe. This also makes it easier for maintainers to remove the recipe without breaking other things.
1. each recipe reuses the `.spago` and `output` folders, so that we only compile `purescript-either` once if 2+ recipes depend on it (either directly or transitively).
1. the root directory's `package.json` file only contains the `devDependencies` for `purescript`, `spago`, and `parcel`. In other words, we don't want this file getting polluted with countless other `npm` dependencies, particularly in cases where one recipe might depend on version A whereas another depends on version B.
[...]

PR #69, opened 2020-06-21
Use pnpm for multi-package monorepo support with faster npm dependency downloads

JordanMartinez:
See [my comment in #56](https://github.com/JordanMartinez/purescript-cookbook/issues/56#issuecomment-647140196) for context.

We might want to try something other than `parcel` (e.g. `esbuild`) to see whether that's what causes `big-integers` to be added to the package.json file when `make BigInts-web` is run.


JordanMartinez:
[Here's the line](https://github.com/JordanMartinez/purescript-cookbook/issues/56#issuecomment-647140196) where `parcel` reports that it can't find the `big-integer` npm dependency. I doubt this because the same code builds find on the `Node.js` backend.

milesfrain:
I'm reproducing this issue locally with:
```
pnpm install
make BigInts-node
make BigInts-web
```

I gave this a shot with parcel V2, and found tt doesn't play nice with pnpm yet https://github.com/parcel-bundler/parcel/issues/3408. It also wanted to add a `babel` devDependency.

We could try esbuild, although that requires some rework because we can't use `index.html` as an entry point https://github.com/evanw/esbuild/issues/31. We'd also need to setup something equivalent to parcel's `--open` flag.

I think the best option for now is to just add all npm dependencies to the top level `package.json` and make a note of the npm dependencies in each recipe's readme. #73 converts this PR to that strategy (I didn't want to push here in case we want to continue the pnpm investigation). We can revisit `parcel v2`, `esbuild`, and perhaps `webpack` if we run into issues with a single `package.json` and need to split them up per recipe. Hopefully some of the previously mentioned bundler issues are sorted-out by that time.

`pnpm` is nice, but let's hold off on making users install another tool unless we really need to. `npm install` in travis is fast enough so far ([~5 seconds](https://travis-ci.com/github/JordanMartinez/purescript-cookbook/builds/172534015#L209))

JordanMartinez:
On my local computer, I tried changing to `esbuild` to see if that fixed the issue. **Unfortunately, it doesn't.** AFAICT, we have three choices:

First, we can figure out why the bundlers (i.e. parcel, esbuild) can't do what Node.js bundler (?) can and determine what to do from there. I'm not sure what the next step of this action would be and it may take a lot of time to figure out.

Second, add a constraint to this repo that all recipes that use `npm` dependencies must use the same versions of those dependencies. If Recipe A through Recipe Y wants to use version 1 and Recipe Z wants to use version 2 of npm dependency, Package, we have a few options. Let's assume Recipe A-Y already exists in the below options:
1. **if someone submitted a PR to add Recipe Z, close Recipe Z's PR.** This seems punitive and would likely discourage contributions, so I'm against it.
1. **if someone submitted a PR to add Recipe Z, force the person to also fix Recipes A-Y.** This seems like it would be fine if the number of recipes are small and it's an easy fix. However, if it requires updating a large number of recipes, it doesn't seem any better than the previous one.
1. in Recipe Z's PR, move Recipe Z into the `broken` folder and open an issue for fixing it later. This seems like a better solution because the cookbook will receive more up-to-date code. 

Third, we could remove the `recipes` folder entirely and push each recipe folder into the root folder. To distinguish recipe folders from non-recipe folders, we could prefix each one with `recipe-`.

milesfrain:
> all recipes that use npm dependencies must use the same versions of those dependencies.

That sounds good.

If we do encounter mismatches, I think preference should go to whatever recipe/library uses the latest npm dependency version.

JordanMartinez:
> I'm not following what advantages this offers.

On second thought, it doesn't fix the problem. Sorry about that.

> If we do encounter mismatches, I think preference should go to whatever recipe/library uses the latest npm dependency version. Older stuff can be moved to broken until the library, package-set, and recipe are updated.

Yeah, that was my thought as well.

> This might be something that should be enforced for package-sets, although I'm not sure if their CI is setup to check for this.

Hmm... that's a good point because how else would the package set maintain it's "they all compile together" guarantee? Is this being discussed on the registry repo yet?

> That sounds good.

I would prefer to find some way to achieve the outcomes I specified above. However, I think this cookbook is "good enough" for the time being. Let's open a new issue for that and further troubleshoot as time permits.

JordanMartinez:
Closing as this PR's main purpose was to test how recipes that require npm dependencies should be handled.

PR #70, opened 2020-06-22
fix %-buildProd target

### LINKED ISSUE #64: Fix broken `%-buildProd` target
See [this explanation](https://github.com/JordanMartinez/purescript-cookbook/pull/62#issue-437418254) for more context.

milesfrain:
Fixes #64 

Interesting makefile behavior with this one.

Shortest stem takes priority... except when:
> a rule whose prerequisites actually exist or are mentioned always takes priority over a rule with prerequisites that must be made by chaining other implicit rules.

And **zero prerequisites** for `recipes/%:` is interpreted as prerequisites that "actually exist or are mentioned".

### DIFF
--- a/makefile
+++ b/makefile
@@ -83,7 +83,10 @@ info:
 # ===== Makefile - Recipe-related Commands =====
 
 # Tests if recipe actually exists.
-recipes/%:
+# The unused dummy lowpriority prerequisite is just to allow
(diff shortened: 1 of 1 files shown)


JordanMartinez:
Wow... that's interesting...

Go ahead and merge.

PR #73, opened 2020-06-22
Add BigInts recipe

milesfrain:
Should this recipe be named `BigInts` or `BigInt`?


JordanMartinez:
This was copied and pasted from the corresponding purescript library's Readme. So, there's a lot more that could exist here (like using FFI to define a constant without using `fromMaybe $ fromNumber 42.0`), so I don't think this is complete. I only submitted it to see how we should handle recipes that use `npm` dependencies.

Let's follow our own principles and open a recipe request that further defines what this recipe would do.

PR #78, opened 2020-06-23
Add constraint: all recipes using npm package foo must use same version of foo

### LINKED ISSUE #76: Add design principle: all recipes using the same `npm` dependency must all use the same version
[In my proposal for how we could deal with `npm` dependnecies](https://github.com/JordanMartinez/purescript-cookbook/pull/69#issuecomment-647755455), @milesfrain and I agreed upon the third approach (paraphrased below):
> Add a constraint to this repo that all recipes that use `npm` dependencies must use the same versions of those dependencies. > Let's say Recipe A through Recipe Y use version 1 and Recipe Z wants to use version 2 of npm dependency, Package. Let's assume Recipe A-Y already exists and someone submitted a PR to add Recipe Z. **In Recipe Z's PR, move Recipe Z into the broken folder and open an issue for fixing it later.** This seems like a better solution because the cookbook will receive more up-to-date code.

So, we need to update Contributing.md to clarify the above principle and the rationale behind it

JordanMartinez:
Fixes #76

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,18 +33,21 @@ Follow these instructions for contributing new recipes. The Goal headers indicat
 
 #### Goal 3: Implement and Submit the Recipe
 
-1. Install needed dependencies via `spago`.
+1. Install needed PureScript dependencies via `spago`.
     - Due to a [bug in Spago (#654)](https://github.com/purescript/spago/issues/654), follow these instructions:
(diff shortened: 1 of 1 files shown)


JordanMartinez:
@milesfrain Could you merge this once CI passes and if I don't get to this before you do?

JordanMartinez:
Thanks! I ran out of time on my lunch break when I was working on this.

PR #80, opened 2020-06-24
Swap package managers: npm -> pnpm

### LINKED ISSUE #74: Which package manager to use: npm, yarn, or pnpm?
Due to #69, I thought it would be better if we discuss this here. See [my comment in #56](https://github.com/JordanMartinez/purescript-cookbook/issues/56#issuecomment-647140196) for more context.

The advantage of using `npm` is that people are already familiar with it and it's already installed on their computer if they are running these recipes in the first place. However, it is the slowest of all three package managers and it does not solve [the doppelganger problem](https://rushjs.io/pages/advanced/npm_doppelgangers/).

The advantage of using `yarn` is that it's the next most familiar package manager aside from `npm` and it's only slightly slower than `pnpm`. However, it doesn't solve the doppelganger problem and I think it's CLI is a bit different than `npm`. So there would be a bit of a learning curve.

The advantage of using `pnpm` is that [it's the fastest of all three package managers](https://pnpm.js.org/en/benchmark) when downloading packages, and it solves the doppelganger problem. It's only disadvantage is that less people know about it. However, it seems that any calls of `npm <command>` can be replaced with `pnpm <command>`. So it's actually more familiar than most people realize. "Installing yet another npm dependency" isn't that big of a deal, IMHO, in this repo because of two reasons. First, we will already be installing numerous npm packages once we get more recipes. Second, all of these packages, including the tools used by this repo (e.g. purescript, spago, parcel), are installed locally, not globally. Finally, this code will likely be wrapped in a `make `command.

While we don't yet have enough recipes to create the doppelganger problem, I am still in favor of using `pnpm`.

JordanMartinez:
Fixes #74

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,11 @@ node_js: 12
 branches:
   only:
   - master
+before_install:
+  - curl -L https://unpkg.com/@pnpm/self-installer | node
+  - pnpm config set store-dir ~/.pnpm-store
 install:
-  - npm install
+  - pnpm install
 script:
   - make testAllCommands
   - diff README.md <(scripts/generateRecipeTable.sh)
@@ -14,3 +17,4 @@ cache:
   directories:
    - output
    - .spago
+   - "~/.pnpm-store"
--- a/README.md
+++ b/README.md
@@ -15,13 +15,9 @@ Install GNU Make, and verify installation with:
 make --version
 ```
 
-Install npm dependencies either globally:
+Install all dependencies locally:
 ```
-npm install -g parcel purescript spago
-```
-Or locally:
-```
-npm install
+make installDeps
 ```
 
 ### Running Recipes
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -4,8 +4,11 @@ node_js: 12
 branches:
   only:
   - master
+before_install:
+  - curl -L https://unpkg.com/@pnpm/self-installer | node
```
milesfrain, line 8: Why not use `npm i pnpm` here too?

JordanMartinez: That does make more sense. I was just following their CI guide, so I'm not sure why they don't recommend that.

JordanMartinez: Fixed in 7218d43af7e5f3b584ad09bb2af3e38c8139337d

PR #93, opened 2020-06-27
Add DiceLog recipe

### LINKED ISSUE #92: Port Elm Numbers example as Node program
**Description of Problem**

How to create a command line prompt and log a random number (die roll) to the console

```
Unique Recipe Name: `DiceLog`
```

**Additional context**

This is intended to be the Node version of [porting Elm's _Numbers_ example](#86).

ptrfrncsmrph:
Fixes #92 
<details><summary><strong>DiceCLI (Node)</strong></summary>

</details>

<details><summary><strong>DiceLog (Node)</strong></summary>

</details>

<details><summary><strong>DiceLog (Browser)</strong></summary>

</details>


milesfrain:
Thanks for the contribution!

Maybe this recipe should be renamed `DiceCLI` because of the interactive component.
I was envisioning the `DiceLog` recipe would be similar to `HelloWorld` in that it just prints a single roll to node or browser console then exits.
We'll also need to tweak CI a bit to indicate that CLI recipes shouldn't be tested for now. An advanced option for later could be to feed in a sequence of keystrokes for testing.

ptrfrncsmrph:
Hm, I wouldn’t be opposed to removing the interactive aspect, would simplify it a bit. If I did that, do you think it would make sense to target the browser as well?

milesfrain:
Good question of whether all recipes that _can_ target the browser _should_ target the browser.

My reasoning is that all recipes with simple logging output should target both (even if nothing is rendered in the browser outside of the console) because that clarifies that there's nothing node-specific (e.g. filesystem) that can't be copied from that recipe into someone's browser project.

So all the code in `DiceLog` would be compatible with the browser, but some of the code in `DiceCLI` wouldn't work in the browser. I think we should have both of these recipes. Feel free to rename / add the other recipe to this PR.

JordanMartinez:
> Good question of whether all recipes that can target the browser should target the browser.
> 
> My reasoning is that all recipes with simple logging output should target both (even if nothing is rendered in the browser outside of the console) because that clarifies that there's nothing node-specific (e.g. filesystem) that can't be copied from that recipe into someone's browser project.
> 
> So all the code in DiceLog would be compatible with the browser, but some of the code in DiceCLI wouldn't work in the browser. I think we should have both of these recipes. Feel free to rename / add the other recipe to this PR.

I've been wondering about this as well. A recipe can currently work on two backends: node and the browser. However, each of these backends could have more specific "targets" (so to speak) within them. For example, the node backend could be a simple run-and-done recipe (e.g. `make RecipeName-node`) or an app with user interaction (maybe `make RecipeName-node-cli`?). Similarly, a browser backend could produce a very simple web app by using Halogen, React, Concur, or plain JavaScript. Thus, we could have two approaches: "one recipe per target" or "multiple targets per recipe."
By splitting each of these into separate recipes (i.e. one recipe per target), we increase CI time and incur maintenance if any one of them needs to be updated (though I don't think that would happen that often). If we combine them into on recipe with multiple targets (i.e. multiple targets per recipe), then it could potentially confuse the end-user which code is being used for a given target and which is used by a separate target. 

I think between these approaches, I would agree with @milesfrain that we should use the "one recipe per target" approach because it only shows the end-user what they need to run it. All other "noise" is removed.

milesfrain:
> "one recipe per target" or "multiple targets per recipe."   ...  I think between these approaches, I would agree with @milesfrain that we should use the "one recipe per target"

To clarify, I'm suggesting `DiceLog` still have multiple targets (node and web).

It's just `DiceCLI` that would have the single node target.

JordanMartinez:
Ah, thanks for clarifying.

JordanMartinez:
@ptrfrncsmrph So, could you do the following?
- rename this recipe to `DiceCLI` and update all usages of the old name to the new one
- rename `nodeSupported.md` to `nodeSupportedSkipCI.md`, so that this recipe will only be compiled, not run when testing CI
- run `make readme` to update the readme

If you want, you could copy your current work, remove the interactivity, name it `DiceLog`, and include that in this PR.

ptrfrncsmrph:
> @ptrfrncsmrph So, could you do the following?
> 
>     * rename this recipe to `DiceCLI` and update all usages of the old name to the new one
> 
>     * rename `nodeSupported.md` to `nodeSupportedSkipCI.md`, so that this recipe will only be compiled, not run when testing CI
> 
>     * run `make readme` to update the readme
> 
> 
> If you want, you could copy your current work, remove the interactivity, name it `DiceLog`, and include that in this PR.

Sounds good! I'll plan on doing that tonight, or sometime this week at the latest.

JordanMartinez:
Could you run `make readme` one more time and push the updated README.md file? Everything else builds fine.

JordanMartinez:
Thanks!

PR #95, opened 2020-06-28
Add window properties

### LINKED ISSUE #87: Access and use browser's `window` and said properties
**Description of Problem**

Show how to use [Window](https://pursuit.purescript.org/packages/purescript-web-html/2.3.0/docs/Web.HTML.Window#t:Window) (e.g. `window >=> document`) and things like that.

```
Unique Recipe Name: WindowProperties
```

**Comparison to other recipes**

This can help one understand how to add an event listener to various actions, but it shouldn't address that specifically

JordanMartinez:
Fixes #87

### DIFF
--- a/makefile
+++ b/makefile
@@ -122,6 +122,7 @@ nodeCompat = $(call recipeDir,$1)/nodeSupported.md
 nodeCompatSkipCI = $(call recipeDir,$1)/nodeSupportedSkipCI.md
 
 # Tests whether recipe can be run on Node.js backend
+.PHONY: %-nodeCompatible
 %-nodeCompatible:
 > @if [ ! -f $(call nodeCompat,$*) ] && [ ! -f $(call nodeCompatSkipCI,$*) ]
 > then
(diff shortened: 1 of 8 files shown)


JordanMartinez:
Sigh... This is a legit recipe, but the build still fails. AFAICT, the issue is that the recipe isn't getting compiled.

JordanMartinez:
`make` is outputting the commands we're using in [the nested if-then-else statements](https://travis-ci.com/github/JordanMartinez/purescript-cookbook/builds/173388252#L297). Locally, I tried adding `@` in front of each one to silence that, but it still gets printed.

milesfrain:
It fails because the `testCI` rule isn't considering dependencies when running the test/build commands.

For example trying to run parcel directly before it's compiled. The `buildWeb` rule handles these dependencies.

It would be even more idomatic to just create all these CI targets and add the prints within the other respective rules. I'll work on that next. This should help with the if-else printing.

The other recipes were being built correctly because their build output was cached from previous CI flow where we built everything. I still think it's better to leave caching enabled.

JordanMartinez:
Turns out the if-else printing gets fixed if I also add a `@` in front of the `$(MAKE)` step. I'm guessing this is because we're using `.ONEHSELL`.

makefile
```diff
@@ -200,23 +200,22 @@ recipes/%/prod/index.html: $(call prodDir,%)
 > then
 >   echo Testing $* on the Node.js backend
 >   spago -x $(call recipeSpago,$*) run --main $(call main,$*)
```
milesfrain, line 202: oops, meant to delete this line too

JordanMartinez: I've fixed this.

PR #97, opened 2020-06-28
Simplify targetsAllCI

milesfrain:
Follow-up to #95


makefile
```diff
@@ -214,5 +214,4 @@ testAllCommands:
 > $(MAKE) HelloWorld-node
 > $(MAKE) HelloWorld-build
 > $(MAKE) HelloWorld-buildWeb
-> $(MAKE) HelloWorld-testCI
```
milesfrain, line 237: We can re-enable these per-recipe `-testCI` rules, but these might not add a lot of value beyond `testAllCI`.

JordanMartinez: Yeah good point.

PR #98, opened 2020-06-29
ReadPrintFileContents - show license instead of readme

milesfrain:
One of the types of CI errors we'll encounter are readme diffs.

Printing a different file during this test helps with visual inspection of the CI logs.


JordanMartinez:
Ah, good point.

Issue #86, opened 2020-06-26
Port Elm Examples


milesfrain:
Proposing we include some (or all) of the [Elm Examples](https://elm-lang.org/examples).

Spork made a [partial start](https://github.com/natefaubion/purescript-spork/tree/master/examples) with this.

These are all web-focused, but some can also have node equivalents. For example, [Numbers](https://elm-lang.org/examples/numbers) doesn't _need_ a roll button. Also, let's rename that one to "Dice".

I'm thinking naming convention could be:
* DiceHalogen - click button to roll
* DiceLog - logs roll to console
* CatGIFsHalogen - this makes more sense for web
* CatGIFsLog - although I guess it could also just log a link that you can click on
* CatGIFsConcur - may demo other frameworks too

It might also be good to make sure each web recipe exists in a single `Main.purs` so we can present TryPureScript links in the table. I'll add a TPS feature that let's us load directly from a file in a repo (so we can bypass the gist upload step).

Added by Jordan with notes on status:
- HTML
    - [x] Hello
    - [x] Groceries
    - [ ] Shapes (blocked until either package set includes Halogen SVG library, we do FFI, or use non-Halogen backend)
- User Input
    - [x] Buttons
    - [x] Text Fields
    - [ ] Forms (blocked until either package set includes Halogen hooks version of Formless, do custom form code, or use non-Halogen backend)
- Random
    - [x] Numbers
    - [x] Cards (note: doesn't use uniform generator when choosing next card)
    - [x] Positions
- HTTP
    - [x] Book
    - [x] Cat GIFs
- Time
    - [x] Time
    - [ ] Clock (blocked until either package set includes Halogen SVG library, we do FFI, or use non-Halogen backend)
- Files
    - [x] Upload
    - [x] Drag-and-Drop
    - [x] Image Previews

paluh:
@milesfrain Do you mean specifically halogen recipes here?

In such a case maybe it should be reflected in the title of the issue so separate tickets could be created for other frameworks.

milesfrain:
No specific framework in mind. Updated list to include a Concur example.

JordanMartinez:
@milesfrain Good idea! 

[...]

PR #105, opened 2020-06-30
Add text fields halogen

JordanMartinez:
Addresses part of #86.

**DO NOT MERGE** - Can't finish until we have a `reverse :: String -> String` function in the package set. We could add it here, but that seems somewhat ridiculous for a simple function like this. (Granted, I realize reversing a string isn't simple due to encoding-related issues).

### DIFF
--- a/recipes/TextFieldsHalogen/src/Main.purs
+++ b/recipes/TextFieldsHalogen/src/Main.purs
@@ -1,4 +1,4 @@
-module HelloHalogen.Main where
+module TextFieldsHalogen.Main where
 
 import Prelude
 
--- a/recipes/TextFieldsHalogen/src/Main.purs
+++ b/recipes/TextFieldsHalogen/src/Main.purs
@@ -31,8 +31,9 @@ hookComponent = Hooks.component \_ _ -> Hooks.do
       [ HH.input
         [ HP.placeholder "Text to reverse"
         , HP.value state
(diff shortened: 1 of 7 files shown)


recipes/TextFieldsHalogen/src/Main.purs
```diff
@@ -0,0 +1,31 @@
...
+  :: forall unusedQuery unusedInput unusedOutput anyMonad
+   . H.Component HH.HTML unusedQuery unusedInput unusedOutput anyMonad
+hookComponent = Hooks.component \_ _ -> Hooks.do
+  state /\ stateIdx <- useState ""
+  Hooks.pure $
+    HH.div_
+      [ HH.input
+        [ HP.placeholder "Text to reverse"
+        , HP.value state
+        , HE.onInput \newValue -> Just $ Hooks.put stateIdx (?reverseString newValue)
```
JordanMartinez, line 31: Somehow, CI built find despite this line. `?reverseString` should have created a compiler error, making the build fail. However, I don't see any such errors [in the build for this recipe](https://travis-ci.com/github/JordanMartinez/purescript-cookbook/builds/173738445#L421). That's a problem...

JordanMartinez: This issue was actually caused by the `web/index.js` file referring to the original recipe I duplicated when I originally started working on this recipe

PR #111, opened 2020-07-01
Ensure index.js points to correct recipe

milesfrain:
Split into three stages:
1. Demonstrate mismatch slipping through CI
1. Add code to catch mismatch in CI - Now caught on [this line](https://travis-ci.com/github/JordanMartinez/purescript-cookbook/builds/173787369#L416)
1. Fix original mismatch

### DIFF
--- a/makefile
+++ b/makefile
@@ -162,13 +162,8 @@ prodDistDir = $(call recipeDir,$1)/prod-dist
 
 # Tests whether recipe can be run on web browser backend
 recipes/%/web:
-> @if [ ! -d $(call webDir,$*) ]
-> then
->   echo
->   echo Recipe $* is not compatible with the web browser backend
->   echo
->   exit 1
-> fi
+> @echo Recipe $* is not compatible with the web browser backend
+> exit 1
 
 .PHONY: %-web
 # Launches recipe in web browser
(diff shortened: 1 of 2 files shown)


makefile
```diff
@@ -162,13 +162,8 @@ prodDistDir = $(call recipeDir,$1)/prod-dist
 
 # Tests whether recipe can be run on web browser backend
 recipes/%/web:
-> @if [ ! -d $(call webDir,$*) ]
```
milesfrain, line 165: Since this rule will only be called if the target needs to be created, the directory check is unnecessary.

Unrelated to this issue, but figured to include it while I was here.

makefile
```diff
@@ -165,17 +166,26 @@ recipes/%/web:
...
+> fi
+
 .PHONY: %-web
 # Launches recipe in web browser
-%-web: $(call recipeDir,%) $(call webDir,%) %-build
+%-web: $(call recipeDir,%) $(call webDir,%) %-indexCheck %-build
 > @echo === Launching $* in the web browser ===
 > parcel $(call webHtml,$*) --out-dir $(call webDistDir,$*) --open
 
 .PHONY: %-buildWeb
```
milesfrain, line 184: I'm not sure if phony targets work with pattern rules. Would be nice to find some documentation on this, although even if they aren't supported (or ignored) they don't seem to be hurting anything.

makefile
JordanMartinez: Yeah, might as well.

JordanMartinez:
Thanks!

PR #115, opened 2020-07-02
Add time halogen

JordanMartinez:
Addresses part of #86


JordanMartinez:
Getting this to display the current time in the current user's time zone would require doing FFI or updating some PureScript libraries. So for now, I'm just going to display UTC time.

PR #118, opened 2020-07-02
Add book halogen

JordanMartinez:
Addresses part of #86

### DIFF
--- a/recipes/BookHalogen/src/Main.purs
+++ b/recipes/BookHalogen/src/Main.purs
@@ -1,0 +1,60 @@
+module BookHalogen.Main where
+
+import Prelude
+
+import Affjax as AX
+import Affjax.ResponseFormat as AXRF
+import Affjax.StatusCode (StatusCode(..))
+import Data.Either (Either(..))
+import Data.HTTP.Method (Method(..))
+import Data.Maybe (Maybe(..))
+import Data.Tuple.Nested ((/\))
+import Effect (Effect)
+import Effect.Aff.Class (class MonadAff, liftAff)
+import Halogen as H
+import Halogen.Aff as HA
+import Halogen.HTML as HH
+import Halogen.Hooks as Hooks
+import Halogen.VDom.Driver (runUI)
+import Network.RemoteData as RD
+
+main :: Effect Unit
+main =
+  HA.runHalogenAff do
+    body <- HA.awaitBody
+    void $ runUI hookComponent Nothing body
+
+hookComponent
+  :: forall unusedQuery unusedInput unusedOutput anyMonad
+   . MonadAff anyMonad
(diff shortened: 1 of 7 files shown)


recipes/BookHalogen/src/Main.purs
```diff
@@ -0,0 +1,60 @@
...
+  :: forall unusedQuery unusedInput unusedOutput anyMonad
+   . MonadAff anyMonad
+  => H.Component HH.HTML unusedQuery unusedInput unusedOutput anyMonad
+hookComponent = Hooks.component \_ _ -> Hooks.do
+  content /\ contentIdx <- Hooks.useState RD.NotAsked
+
+  Hooks.useLifecycleEffect do
+    Hooks.put contentIdx RD.Loading
+    errorOrResponse <- liftAff $ AX.request $ AX.defaultRequest
+      { url = "https://elm-lang.org/assets/public-opinion.txt"
```
milesfrain, line 37: I don't think we should grab this from elm-lang.org.

We could go with Project Gutenberg instead, and use a mirror to obey their [TOS](http://www.gutenberg.org/wiki/Gutenberg:Terms_of_Use).

Here's a link to "Public Opinion" (649 kB): https://gutenberg.pglaf.org/6/4/5/6456/6456.txt

I think it would be nice to go a bit easier on the mirror with a smaller text, such as "Ion by Plato" (55 kB): https://gutenberg.pglaf.org/1/6/3/1635/1635.txt

JordanMartinez: I tried using both links above and the recipe renders the "I couldn't load the page" error message. When I added `printError` to the message, the message said:
> There was a problem making the request: AJAX request failed: GET https://gutenberg.pglaf.org/6/4/5/6456/6456.txt

JordanMartinez: I'm find with using the smaller text, btw.

milesfrain: Looks like an ssl issue.

```
> curl https://gutenberg.pglaf.org/1/6/3/1635/1635.txt | wc
curl: (60) SSL certificate problem: unable to get local issuer certificate
More details here: https://curl.haxx.se/docs/sslcerts.html
curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
      0       0       0

> curl http://gutenberg.pglaf.org/1/6/3/1635/1635.txt | wc 
   1201    9539   55890
```

So two options are:
http://gutenberg.pglaf.org/1/6/3/1635/1635.txt
http://aleph.gutenberg.org/1/6/3/1635/1635.txt

PR #119, opened 2020-07-02
Add cat gifs halogen

JordanMartinez:
Addresses part of #86

One problem. The URL used to get a cat GIF from giphy uses an API key. This is likely Elm's API key. We should get our own and use that as the link rather than theirs.


JordanMartinez:
Also, should this recipe be renamed to `CatGifsHalogen`? I've mistyped the current name due to typing `GIF` followed by `s` followed by `H`.

milesfrain:
That key is the "public beta key" so it's fine to use https://giphy.api-docs.io/1.0/welcome/access-and-api-keys

Leaning towards `Gif` over `GIF` due to ease of typing.

JordanMartinez:
Ah, thanks for figuring that out!

Recipe has been renamed to `CatGifsHalogen`.

paulyoung:
> The URL used to get a cat GIF from giphy uses an API key

Busted 😄 https://mobile.twitter.com/havemysecrets/status/1278682391108861953

milesfrain:
They can have that secret

JordanMartinez:
Nice! Free publicity!

PR #121, opened 2020-07-03
Cut down on recipe request paperwork

milesfrain:
I feel like a simpler recipe request form will make this repo more inviting.

We can always add more requirements if we notice a pattern of missing content in recipe requests.


JordanMartinez:
Makes sense!

PR #127, opened 2020-07-04
add buildAll command

milesfrain:
Compiling the PS code for all recipes seems to help with the VSCode IDE when editing, although there are still some additional configurations to sort-out for dealing with this non-traditional workspace.


JordanMartinez:
Thanks! I think it helps because all dependencies are compiled and then picked up by `purs ide` or something.

PR #130, opened 2020-07-04
Add CardsReact

ptrfrncsmrph:
Partially addresses #86

### DIFF
--- a/recipes/CardsReact/README.md
+++ b/recipes/CardsReact/README.md
@@ -6,7 +6,7 @@ A React port of the ["Random - Cards" Elm Example](https://elm-lang.org/examples
 
 ### Browser
 
-If the user clicks the "Draw" button, a random card will be displayed. **Note:** the Elm example uses a uniform generator whereas this example does not because PureScript does not have an out-of-box uniform generator.
+If the user clicks the "Draw" button, a random card will be drawn from a deck of 13 spades and displayed.
 
 ## Dependencies Used:
 
--- a/recipes/CardsReact/README.md
+++ b/recipes/CardsReact/README.md
@@ -1,0 +1,14 @@
+# CardsReact
+
+A React port of the ["Random - Cards" Elm Example](https://elm-lang.org/examples/cards).
+
+## Expected Behavior:
+
+### Browser
+
+If the user clicks the "Draw" button, a random card will be displayed. **Note:** the Elm example uses a uniform generator whereas this example does not because PureScript does not have an out-of-box uniform generator.
+
+## Dependencies Used:
+
+[react](https://www.npmjs.com/package/react)
+[react-dom](https://www.npmjs.com/package/react-dom)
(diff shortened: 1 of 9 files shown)


milesfrain:
There are a few options to address:
> **Note:** the Elm example uses a uniform generator whereas this example does not because PureScript does not have an out-of-box uniform generator.

The first possible fix is to just make a helper function that randomly selects a value from a non-empty array of Cards. Then we can get rid of the `intToCard` block.

Another fix that more closely matches Elm's generator is to use a combo of [`elements`](https://pursuit.purescript.org/packages/purescript-quickcheck/6.1.0/docs/Test.QuickCheck.Gen#v:elements) and [`runGen`](https://pursuit.purescript.org/packages/purescript-quickcheck/6.1.0/docs/Test.QuickCheck.Gen#v:runGen). This just requires tracking the resulting `Tuple a GenState`. I think that can just be a single piece of react state.

This tuple tracking is also what Elm is doing behind the scenes with [`generate`](https://github.com/elm/random/blob/c1c9da4d861363cee1c93382d2687880279ed0dd/src/Random.elm#L860) and `onEffects`. Elm tracks this tuple as `(value, newSeed)`.

This feedback also applies to the Halogen version of this recipe.

recipes/CardsReact/README.md
```diff
@@ -0,0 +1,14 @@
+# CardsReact
+
+A React port of the ["Random - Cards" Elm Example](https://elm-lang.org/examples/cards).
+
+## Expected Behavior:
+
+### Browser
+
+If the user clicks the "Draw" button, a random card will be displayed. **Note:** the Elm example uses a uniform generator whereas this example does not because PureScript does not have an out-of-box uniform generator.
```
milesfrain, line 9: Also would be good to mention that only Spades are drawn.
> If the user clicks the "Draw" button, a random card will be drawn from a deck of 13 spades and displayed.

ptrfrncsmrph:
> Another fix that more closely matches Elm's generator is to use a combo of [`elements`](https://pursuit.purescript.org/packages/purescript-quickcheck/6.1.0/docs/Test.QuickCheck.Gen#v:elements) and [`runGen`](https://pursuit.purescript.org/packages/purescript-quickcheck/6.1.0/docs/Test.QuickCheck.Gen#v:runGen). This just requires tracking the resulting `Tuple a GenState` (where `a` is `Card`). I think that can just be a single piece of react state.

👍 I went with this approach.

A bit confused by this build error
```
 ERROR  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up-to-date with package.json

The command "pnpm install" failed and exited with 1 during .
```
I've only edited `README.md`, `Main.purs` & `spago.dhall` since it last passed, and haven't added any `npm` dependencies.

JordanMartinez:
Personally, I think `Gen` should be moved outside of `quickcheck`. There are a number of times I would like to generate random values but not use the testing framework.
Still, we can accept this and work on that another time.

Build is failing with this error:
```
$ pnpm install

Lockfile is up-to-date, resolution step is skipped

 ERROR  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up-to-date with package.json

The command "pnpm install" failed and exited with 1 during .
```
I believe I encountered this before and solved it by deleting the `pnpm-lock.yaml` file and then reinstalling.

JordanMartinez:
Thanks!

milesfrain:
> Personally, I think `Gen` should be moved outside of `quickcheck`.

I was thinking the same.

PR #133, opened 2020-07-04
Add CatGifsReact

ptrfrncsmrph:
Partially addresses #86

### DIFF
--- a/recipes/CatGifsReact/src/Main.purs
+++ b/recipes/CatGifsReact/src/Main.purs
@@ -1,0 +1,92 @@
+module CatGifsReact.Main where
+
+import Prelude
+import Affjax (printError)
+import Affjax as Affjax
+import Affjax.ResponseFormat as ResponseFormat
+import Data.Argonaut.Decode (decodeJson)
+import Data.Argonaut.Decode.Combinators ((.:))
+import Data.Array (singleton)
+import Data.Bifunctor (lmap)
+import Data.Either (Either(..))
+import Data.Maybe (Maybe(..))
+import Effect (Effect)
+import Effect.Aff (Aff)
+import Effect.Class.Console (log)
+import Effect.Exception (throw)
+import React.Basic.DOM (css, render)
+import React.Basic.DOM as R
+import React.Basic.Events (handler_)
+import React.Basic.Hooks (Component, component, useEffectOnce, (/\))
+import React.Basic.Hooks as React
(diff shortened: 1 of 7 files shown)


recipes/CatGifsReact/src/Main.purs
```diff
@@ -0,0 +1,92 @@
...
+
+data GifAction
+  = MorePlease
+  | GotGif (Either String String)
+
+reducer :: AffReducer GifState GifAction
+reducer state = case _ of
+  MorePlease ->
+    { state: Loading
+    , effects: [ getRandomCatGif <#> GotGif >>> singleton ]
```
ptrfrncsmrph, line 41: This is my first time using `useAffReducer`, and the API is a little bit odd, but I understand that it's also close to the Elm way of doing things so it might be appropriate here (I'm also not familiar with Elm). In any case this `getRandomCatGif <#> GotGif >>> singleton` is because `effects` needs to be of type `Array (Aff (Array GifAction))`. There might be a cleaner way of doing this, but I thought it would make sense to read it from left to right: `getRandomCatGif` and then put the result of that into `GotGif` action, and then put that into an array.

recipes/CatGifsReact/src/Main.purs
```diff
@@ -0,0 +1,92 @@
...
+
+getRandomCatGif :: Aff (Either String String)
+getRandomCatGif = do
+  log "Foo"
+  response <-
+    Affjax.get
+      ResponseFormat.json
+      "https://api.giphy.com/v1/gifs/random?api_key=dc6zaTOxFJmzC&tag=cat"
+  pure do
+    { body } <- lmap printError response
```
ptrfrncsmrph, line 89: The `printError` here is just so the error types line up between the `Affjax.get` and the `decodeJson` result. That error is ultimately ignored, so an alternative is using `hush` and converting the type to `Aff (Maybe String)`.

recipes/CatGifsReact/src/Main.purs
JordanMartinez: I'm not familiar enough with the precedence of `<#>` vs `>>>`. So, it's not obvious to me which one is actually "run" first. What if this was written as
`(singleton <<< GotGif) <$> getRandomCatGif`?

JordanMartinez:
LGTM. I would prefer the change mentioned above, but I'm not against leaving it as that. Precedence is something I should become more familiar with.

JordanMartinez:
Thanks!

PR #135, opened 2020-07-06
Make version note CI fixup

milesfrain:


JordanMartinez:
@milesfrain I restarted the build and it still failed. Looks like the issue may have something to do with dependencies getting updated during CI's build? Somewhat related...? https://stackoverflow.com/questions/55284678/how-to-use-my-old-shrinkwrap-yaml-with-pnpm-3-0

PR #136, opened 2020-07-06
Dummy PR

milesfrain:
Just for testing CI and pnpm lockfile issue


JordanMartinez:
Build is working. I guess we can just ignore the `pnpm-lock.yaml` file in the `.gitignore` file until we find a better solution?

JordanMartinez:
I've updated `.gitignore` to ignore that file and pushed the change to master.

PR #137, opened 2020-07-06
Update recipe names to use "recipe suffix" to help clarify what they do

### LINKED ISSUE #123: Thoughts on standardizing "targets"?
So far, we've been unofficially following a naming convention. Depending on which library/interaction is used, a suffix is being added to the recipe name:
- `*Halogen` = Halogen is being used
- `*React` = React is being used
- `*CLI` = Node is being used and it involves user interaction
- `*Node` = Node is being used and no user interaction occurs

Would we like to keep using this convention? If so, perhaps there's a few others we can add

- `*Js` = plain JavaScript is being used (i.e. no framework like Halogen/React/Concur)
- `*Concur` = Concur is being used (if and when we get a PR for it)

JordanMartinez:
Fixes #123


JordanMartinez:
Note: this is still a WIP. I still need to update a few more recipes' names

JordanMartinez:
Hm... still the same issue as before...

JordanMartinez:
Looks like this will build now. Readme diff led to a CI build failure, so I've fixed that in latest commit.

PR #142, opened 2020-07-07
Fix CI

JordanMartinez:


JordanMartinez:
Fixed the `pnpm` issue by wiping `node_modules` and reinstalling. Not sure why that works, but this appears to be a cache issue...?

Regardless, build now fails because we don't have `object-assign` as a dependency.

JordanMartinez:
@milesfrain Do you have any ideas for how we can fix this build issue? All the other work I would have done today was blocked by this.

milesfrain:
Merging this without `pnpm`. I want to run a test with spago and `npm` cached for master to check CI timing. We can re-introduce `pnpm` later.

milesfrain:
It looks like travis plays nicer with `npm` than `pnpm`, and it's even slightly faster with `npm`.

- 2m38s for #133 with `pnpm`
- 2m11s for #143 with `npm`

So proposing we stick with `npm` for simplicity and speed.

JordanMartinez:
> So proposing we stick with npm for simplicity and speed.

I'm good with that. We're not using the monorepo support of pnpm because it won't work with our current situation and pnpm caused quite a huge delay. We should update the instructions that are affected by this.

PR #144, opened 2020-07-07
add info for Nix-users to install dependencies.

danielstaleiny:
- readme install information
  - How to remove spago, purescript dependencies with 1 line
  command. (unix only)
- create pinned purs 0.13.6 dependencies with spago and the rest. shell.nix

Let me know if you want to pin higher version of the purs, I picked 0.13.6 because I saw that in packages.dhall

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,13 @@ Install all dependencies locally:
 make installDeps
 ```
 
+Install for Nix users:
+```
+# Remove purescript and spago and removing traling , from package.json
+sed --in-place '/purescript\|spago/d' ./package.json && sed --in-place '$!N;s/,\n  }/\n  }/;P;D' package.json
+nix-shell
+```
+
 ### Running Recipes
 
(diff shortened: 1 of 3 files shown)


JordanMartinez:
@danielstaleiny Thanks for the PR. To ease maintenance burden, we generate the readme by combining the `scripts/ReadmeContent.md` with the recipe ToC that we produce using bash. 

Please make the following changes, so that CI builds:
- move your ReadMe.md updates to scripts/ReadmeContent.md
- run `make readme`

danielstaleiny:
Done, I override the previous changes, so the history stay clean. 
and thanks for the info, I was a bit confused why it failed with readme complaining.

JordanMartinez:
Thanks!

milesfrain:
> Done, I override the previous changes, so the history stay clean.

No need to force push. We use "squash and merge" on PRs which creates a clean linear history of 1 commit per PR (even if the PR contains lots of commits).

PR #151, opened 2020-07-07
Add FileUploadHalogenHooks

JordanMartinez:
Addresses part of #86

### DIFF
--- a/recipes/FileUploadHalogenHooks/README.md
+++ b/recipes/FileUploadHalogenHooks/README.md
@@ -1,0 +1,9 @@
+# FileUploadHalogenHooks
+
+A Halogen port of the ["Files - Upload" Elm Example](https://elm-lang.org/examples).
+
+## Expected Behavior:
+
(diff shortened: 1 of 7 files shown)


recipes/FileUploadHalogenHooks/README.md
```diff
@@ -0,0 +1,9 @@
+# FileUploadHalogenHooks
+
+A Halogen port of the ["Files - Upload" Elm Example](https://elm-lang.org/examples).
```
milesfrain, line 3: Thoughts on linking directly to the example?
https://elm-lang.org/examples/upload

JordanMartinez: Initially, I was hesitant on doing that because one can't find the rest of the Elm examples if we do that. But I realized there's a "More examples `here`" in the upper-left corner, so that's not true...

Sure, let's update the Readmes to point directly to the code.

PR #152, opened 2020-07-08
Add DragAndDropHalogenHooks

JordanMartinez:
Addresses part of #86

This one took a while to figure out because Elm is doing some unsafe things behind the scenes. I decided to use a different approach.


JordanMartinez:
This PR added a few [compiler warnings due to unused imports](https://travis-ci.com/github/JordanMartinez/purescript-cookbook/builds/174733967#L596). I'll fix this in a separate PR.

PR #155, opened 2020-07-08
DragDropHalogen - Simplify input in label

milesfrain:
Original:
```
label
  div (button style)
input
span
```

New:
```
label (button style)
  input
span
```

### DIFF
--- a/recipes/DragAndDropHalogenHooks/src/Main.purs
+++ b/recipes/DragAndDropHalogenHooks/src/Main.purs
@@ -2,7 +2,7 @@ module DragAndDropHalogenHooks.Main where
 
 import Prelude hiding (top)
 
-import CSS (alignItems, border, borderRadius, color, column, dashed, display, displayNone, flex, flexDirection, gray, height, justifyContent, margin, padding, purple, px, solid, width)
+import CSS (alignItems, backgroundColor, border, borderRadius, color, column, dashed, display, displayNone, flex, flexDirection, gray, height, justifyContent, lightblue, margin, padding, purple, px, solid, width)
 import CSS as CSS
 import CSS.Common (center)
 import DOM.HTML.Indexed.InputType (InputType(..))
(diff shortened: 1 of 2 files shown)


recipes/DragAndDropHalogenHooks/src/Main.purs
```diff
@@ -69,40 +69,52 @@ hookComponent = Hooks.component \_ _ -> Hooks.do
           mbFileList = DataTransfer.files $ dataTransfer e
           fileArray = maybe [] FileList.items mbFileList
         Hooks.put filesIdx fileArray
+        Hooks.put hoverIdx false
```
milesfrain, line 72: Clears highlighting after file drop.

recipes/DragAndDropHalogenHooks/web/index.css
```diff
@@ -1,4 +1,3 @@
 .otherCssNotInPurescript-Css {
-  background-color: lightblue;
   cursor: pointer;
```
milesfrain, line 2: We could get rid of this last CSS workaround once https://github.com/purescript-contrib/purescript-css/pull/94 is merged.
Then we'll be able to host this recipe on TPS.

recipes/DragAndDropHalogenHooks/src/Main.purs
```diff
@@ -69,40 +71,53 @@ hookComponent = Hooks.component \_ _ -> Hooks.do
...
-    -- 2. add it to the DOM
-    -- 3. create a mouse event
-    -- 4. dispatch the mouse event to the input element
-    -- 5. (implication) file dialogue appears
-    -- 6. user selects a file
-    -- 7. input event handler runs a callback using user's selected file
-    -- 8. input element is removed from DOM
-    --
-    -- The approach used below is based on this SO answer:
-    -- https://stackoverflow.com/a/47094148
```
JordanMartinez, line 84: If you're going to delete this, maybe it should be moved to the recipe's Readme?

recipes/DragAndDropHalogenHooks/src/Main.purs
JordanMartinez: Good point

recipes/DragAndDropHalogenHooks/web/index.css
JordanMartinez: Sounds good!

recipes/DragAndDropHalogenHooks/src/Main.purs
milesfrain: I just moved this section to the bottom of the file

JordanMartinez: Ah, I hadn't scrolled that far... Thanks for clarifyin!

PR #161, opened 2020-07-08
Add DebuggingLog

### LINKED ISSUE #116: Debugging
**Description of Problem**

Debugging can be non-obvious to PureScript beginners. First, you can't insert log statements in pure code, and second, you can only log `String` values, so if your data doesn't have a `Show` instance you're out of luck.

The `Debug.Trace` module has plenty of helper functions that help with this. The two I use the most are `spy` and `traceM`, but on the whole this package lets you get "print debugging" when you temporarily need it. Even better, these functions carry a compiler warning that helps you remember to remove the statements before say opening a pull request.

A useful recipe might show how to insert debugging statements into pure (and/or effectful) code using various functions from the `Debug.Trace` modules.

```
Unique Recipe Name: Debug
```

**Comparison to other recipes**

Other recipes use logging as a proof that they work, but none show how to use `Debug.Trace` to log pure code or insert temporary statements. This module is a simple but immediately useful tool to know about.

JordanMartinez:
Fixes #116 

This doesn't cover `trace` yet, though it should. Please give feedback.

### DIFF
--- a/recipes/DebuggingLog/README.md
+++ b/recipes/DebuggingLog/README.md
@@ -1,0 +1,9 @@
+# DebuggingLog
+
+This recipe shows how to do print-debugging using the `Debug` module's `spy` and `traceM` functions.
+
+## Expected Behavior:
+
+Does console-printing-based debugging in various contexts to show the pros/cons of each approach and how reliable/unpredictable they can be.
+
+For the browser, make sure to open the console with dev tools first, then reload/refresh the page.
--- a/recipes/DebuggingLog/README.md
+++ b/recipes/DebuggingLog/README.md
@@ -1,6 +1,6 @@
 # DebuggingLog
 
-This recipe shows how to do print-debugging using the `Debug` module's `spy` and `traceM` functions.
+This recipe shows how to do print-debugging using the `Debug` module's `spy` and `traceM` functions. The compiler will emit warnings to remind you to remove these debug functions before you ship production code.
 
 ## Expected Behavior:
 
--- a/recipes/DebuggingLog/src/Main.purs
+++ b/recipes/DebuggingLog/src/Main.purs
@@ -32,7 +32,6 @@ main = do
 
   usingTraceMInEffect
   usingTraceMInST
-  usingTraceMInAff
   usingTraceMInAffRace
 
 usingSpy :: Effect Unit
@@ -81,13 +80,6 @@ usingTraceMInST = do
         traceM four
         ST.write (four + 2) localReference
   log $ "Local reference value is: " <> show localMutationResult
-
-usingTraceMInAff :: Effect Unit
-usingTraceMInAff = do
-  launchAff_ do
-    pure unit
-    traceM 5
-    pure unit
 
 usingTraceMInAffRace :: Effect Unit
 usingTraceMInAffRace = do
--- a/recipes/DebuggingLog/src/Main.purs
+++ b/recipes/DebuggingLog/src/Main.purs
@@ -32,6 +32,8 @@ main = do
 
   usingTraceM
 
+  compareSpyAndTraceM
+
 usingSpy :: Effect Unit
 usingSpy = do
   log "usingSpy"
@@ -82,3 +84,9 @@ usingTraceM = do
       ST.write (four + 2) localReference
 
   log $ "Local reference value is: " <> show (ST.run localMutationComputation)
+
+compareSpyAndTraceM :: Effect Unit
+compareSpyAndTraceM = do
+  let x = 5
+      _ = spy "x" x
+  traceM x
--- a/recipes/DebuggingLog/src/Main.purs
+++ b/recipes/DebuggingLog/src/Main.purs
@@ -69,7 +69,7 @@ usingSpy = do
 usingTraceMInEffect :: Effect Unit
 usingTraceMInEffect = do
   log "usingTraceMInEffect"
-  log "But what if yo udon't want to depend on `console` to do temporary \
+  log "But what if you don't want to depend on `console` to do temporary \
       \printing-based debugging? We can use traceM instead."
   traceM "logging to the console without using `log`"
 
--- a/recipes/DebuggingLog/src/Main.purs
+++ b/recipes/DebuggingLog/src/Main.purs
@@ -31,7 +31,6 @@ main = do
   usingSpy
 
   usingTraceM
-  usingTraceMInAffRace
 
 usingSpy :: Effect Unit
 usingSpy = do
@@ -76,34 +75,3 @@ usingTraceM = do
         traceM four
         ST.write (four + 2) localReference
   log $ "Local reference value is: " <> show localMutationResult
-
-usingTraceMInAffRace :: Effect Unit
-usingTraceMInAffRace = do
-  launchAff_ do
-    parSequence [ runXTimes 4 $ genValueTraceMNo $ spy "which fiber id" 1
-                , runXTimes 4 $ genValueTraceMYes 2
-                , runXTimes 4 $ genValueTraceMYes 3
-                , runXTimes 4 $ genValueTraceMYes 4
-                , runXTimes 4 $ genValueTraceMNo 5
-                , runXTimes 4 $ genValueSpy 5
-                ]
-
-  where
-    runXTimes 1 comp = comp
-    runXTimes n comp = comp *> runXTimes (n - 1) comp
-
-    genValueTraceMNo _ = do
-      randomValue <- liftEffect $ randomInt 1 10
-      delay $ Milliseconds $ toNumber randomValue
-      pure unit
-
-    genValueTraceMYes i = do
-      randomValue <- liftEffect $ randomInt 1 10
-      delay $ Milliseconds $ toNumber randomValue
-      traceM $ "Fiber: " <> show i <> " - random value: " <> show randomValue
-      pure unit
-
-    genValueSpy i = do
-      randomValue <- liftEffect $ randomInt 1 10
-      delay $ Milliseconds $ toNumber $ spy "(spy) random value" randomValue
-      pure unit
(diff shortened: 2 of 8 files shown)


JordanMartinez:
ping @thomashoneyman

recipes/DebuggingLog/src/Main.purs
```diff
@@ -0,0 +1,116 @@
...
+
+  log $ "Thus, spying is an effective way of quickly adding debugging where \
+        \you need it without affecting any other part of your code. \
+        \Note: you should not do this in production code. Use a proper logger \
+        \in production code."
+
+usingTraceMInEffect :: Effect Unit
+usingTraceMInEffect = do
+  log "usingTraceMInEffect"
+  log "But what if yo udon't want to depend on `console` to do temporary \
```
milesfrain, line 72: "you" typo

recipes/DebuggingLog/src/Main.purs
```diff
@@ -0,0 +1,116 @@
...
+        ST.write (four + 2) localReference
+  log $ "Local reference value is: " <> show localMutationResult
+
+usingTraceMInAff :: Effect Unit
+usingTraceMInAff = do
+  launchAff_ do
+    pure unit
+    traceM 5
+    pure unit
+
```
milesfrain, line 92: Some comments on what to expect here would be helpful

recipes/DebuggingLog/src/Main.purs
```diff
@@ -0,0 +1,116 @@
...
+  let localMutationResult = ST.run do
+        localReference <- ST.new 0
+        traceM localReference
+        four <- ST.modify (_ + 4) localReference
+        traceM four
+        ST.write (four + 2) localReference
+  log $ "Local reference value is: " <> show localMutationResult
+
+usingTraceMInAff :: Effect Unit
+usingTraceMInAff = do
```
thomashoneyman, line 87: `traceM` ought to be valid in any monad, and it's used in the `launchAff_` above, so I'm not sure it's necessary to repeat again here.

recipes/DebuggingLog/src/Main.purs
```diff
@@ -0,0 +1,116 @@
...
+  let
+    x = 5
+    y = spy "y" 8
+    adt = spy "adt" $ MyADT 1 (Tuple 4 ["a", "b"]) { foo: "foo value" }
+    function = spy "function" $ \intValue -> show $ 4 + intValue
+
+    -- quick way of debugging something without showing it or using a
+    -- variable name
+    _ = spy "debug for me what x + y is" $ x + y
+    arrayOfStrings = spy "debug some array" $ map show [1, 2, 3, 4, 5]
```
thomashoneyman, line 55: One other place it's common to use `spy` is, for example, in a render function:

```purs
render state =
  HH.div
    [ ]
    [ HH.text state.someValue ]
```

becomes

```purs
...
    [ HH.text $ spy "state.someValue" state.someValue ]
```

This makes `spy` really convenient to drop into arbitrary code. You simply add `spy "label"` in front of whatever value it is you had before, and you get the logging without any other change to your code.

Another example: let's say you've written a little algorithm with a loop and you want to spy at each step:

```purs
fact :: Int -> Int -> Int
fact 0 acc = acc
fact n acc = fact (spy "n" (n - 1)) (spy "acc" (acc * n))
```

Being able to insert `spy` inline is really nice. I wanted to call it out because you can't do this with `traceM`.

recipes/DebuggingLog/README.md
```diff
@@ -0,0 +1,9 @@
+# DebuggingLog
+
+This recipe shows how to do print-debugging using the `Debug` module's `spy` and `traceM` functions.
+
+## Expected Behavior:
+
+Does console-printing-based debugging in various contexts to show the pros/cons of each approach and how reliable/unpredictable they can be.
```
thomashoneyman, line 7: You might also want to include that these functions include a compiler warning, which helps ensure this code gets removed in say the code review process rather than make it to production.

It's similar to why some folks define a custom `todo` function like this:

```purs
todo
  :: forall sym a
   . IsSymbol sym
  => Warn (Beside (Text "TODO: ") (Text sym))
  => SProxy sym
  -> a
todo sym =
  unsafeThrowException $ error $ "TODO: " <> reflectSymbol sym
```

which allows you to make a TODO comment that also includes a compiler warning so you don't forget it.

recipes/DebuggingLog/src/Main.purs
JordanMartinez: Thanks

recipes/DebuggingLog/README.md
JordanMartinez: Good point. I've updated the readme to include a note about this, so that it shows up in the recipe ToC, too.

recipes/DebuggingLog/src/Main.purs
JordanMartinez: Added the `fact` example and updated the `traceM` usage to include two `spy`s to show how that works.

recipes/DebuggingLog/src/Main.purs
JordanMartinez: Yeah.... I wanted to distinguish this from the racing version of `Aff`. That's probably not needed.

recipes/DebuggingLog/src/Main.purs
JordanMartinez: I believe `traceM` in Haskell doesn't happen in a predictable way. I was trying to do the same here. On second thought, I'm not sure what the problems are with using `traceM` in Haskell and whether those problems still arise in PureScript. Maybe this should be removed entirely?

JordanMartinez:
Any examples you two can think of for `trace`?

thomashoneyman:
`trace` delays evaluating the value it's going to show until the message has been printed, so I suppose it would be used when you are tracing something that logs to the console and you want your trace message to happen first. I haven't used this myself, though.

recipes/DebuggingLog/src/Main.purs
```diff
@@ -0,0 +1,129 @@
...
+  traceM "logging to the console without using `log`"
+
+usingTraceMInST :: Effect Unit
+usingTraceMInST = do
+  let localMutationResult = ST.run do
+        localReference <- ST.new 0
+        traceM localReference
+        four <- ST.modify (_ + 4) localReference
+        traceM four
+        ST.write (four + 2) localReference
```
milesfrain, line 90: I'm not seeing the advantage of `traceM` over `spy` in this case (assuming there's a "tagless" version of `spy` available).
```purs
usingTraceMInST = do
  let localMutationResult = ST.run do
        localReference <- ST.new 0
        four <- ST.modify (_ + 4) $ spy localReference
        ST.write ((spy four) + 2) localReference
```

Maybe the benefit is convenience in controling the order that values are printed in and being able to easily add/remove these debug prints.

milesfrain:
Should we improve the `debug` API?

Thinking something like:
```purs
-- Uses JS console.log to print contents
spy :: a -> a
-- Uses Show instance to convert contents to string
spyShow :: Show a => a -> a
-- Prepends a tag (current spy behavior)
spyTag :: String -> a -> a
spyTagShow :: Show a => String -> a -> a

-- Wait to evaluate expression
-- Not sure if this only makes sense with `Tag`
spyThunk :: (Unit -> a) -> a
-- ... other combinations with Tag and Show

-- Monodic version
-- Wondering if this could be relaxed to Applicative (like haskell's traceM).
spyM :: Monad m => a -> m Unit
-- ... other combinations with Thunk, Tag, and Show

-- Stack dump
-- Uses JS console.trace() to log stack
-- Similar to Haskell's traceStack
spyStack :: a -> a
-- ... other combinations with M, Thunk, Tag, and Show

-- Reflecting version
-- Uses reflected expression as tag
-- Not sure if this is possible
spyReflect :: a -> a
-- ... other combinations with Stack, M, Thunk, Tag, and Show

-- This would be great to include
todo :: a
```

**Additional details on `spyReflect`**

Imagine writing:
```purs
fact :: Int -> Int -> Int
fact 0 acc = spyReflect acc 
fact n acc = fact (spyReflect $ n - 1) (spyReflect $ acc * n)
```
To get this behavior:
```purs
fact :: Int -> Int -> Int
fact 0 acc = spy "acc" acc 
fact n acc = fact (spy "n - 1" $ n - 1) (spy "acc * n" $ acc * n)
```
This is really convenient for debugging in other languages:
```jl
julia> x = 5
5

julia> @show (x + 3)
x + 3 = 8
8

julia> @info "tag" (x + 3)
┌ Info: tag
└   x + 3 = 8
```

**Other questions:**
* Thoughts on `spy` vs `trace` naming?
* Are there other ways to gain all these behaviors without writing 2^n functions? Looks like 64 would be needed.
* `Monad` vs `Applicative` for `traceM`/`spyM`?
* Any other inspiration to gain from [Haskell](https://hackage.haskell.org/package/base-4.14.0.0/docs/Debug-Trace.html) or [Elm](https://package.elm-lang.org/packages/elm/core/latest/Debug) debug logging?
* Which `todo` implementation to use? This [PR](https://github.com/garyb/purescript-debug/pull/24) or this [snippet](https://github.com/JordanMartinez/purescript-cookbook/pull/161#discussion_r451940206)?

recipes/DebuggingLog/src/Main.purs
JordanMartinez: Honestly, I'm not sure why `traceM` exists when we have `spy`. Maybe because of the following:
```
let _ = spy "foo" foo
-- vs
traceM foo
```

JordanMartinez:
I'm leaning against adding `trace` here for the time being. I'd like to get this merged and then add improvements later.

JordanMartinez:
If you think the current state of this is fine, I'll merge tomorrow. Build was failing because I hadn't regenerated the readme.

JordanMartinez:
If you think this should be modified in some way, open an issue for it and I can submit another PR to fix that.

milesfrain:
Moved debug API discussion to garyb/purescript-debug#25

Issue #138, opened 2020-07-07
Port all Halogen examples to HalogenHooks


milesfrain:
- [ ] AceHalogenHooks #120
- [x] BasicHalogenHooks                                                                                  
- [x] ComponentsHalogenHooks                                                                             
- [x] ComponentsInputsHalogenHooks                                                                       
- [x] ComponentsMultitypeHalogenHooks                                                                    
- [x] DriverIoHalogenHooks                                                                               
- [x] DriverRoutingHalogenHooks                                                                          
- [x] DriverWebsocketsHalogenHooks                                                                       
- [ ] EffectsAffAjaxHalogenHooks                                                                         
- [ ] EffectsEffectRandomHalogenHooks                                                                    
- [ ] HigherOrderComponentsHalogenHooks                                                                  
- [x] InterpretHalogenHooks                                                                              
- [x] KeyboardInputHalogenHooks                                                                          
- [x] LifecycleHalogenHooks

https://github.com/purescript-halogen/purescript-halogen/tree/master/examples

JordanMartinez:
I'm claiming this one next.

JordanMartinez:
`ace` isn't in the package set :-/

JordanMartinez:
I'm not sure whether the following recipes should be implemented based on their readme's content. Thoughts?
- `effects-aff-ajax`: "demonstrates how to make an AJAX request in a component `eval` function." Isn't this already shown via `ImagePreviewsHalogenHooks`?
- `effects-effect-random`: "demonstrates how to use an `Effect` function in a component `eval` function." Isn't this already shown in various recipes here already? The example they use is generating a random number. We've already got recipes for doing that...
- `higher-order-components`: "illustrates how to define and use a higher-order component". Hooks kind of make higher-order components pointless now because one can just define a hook (e.g. halogen-select) that can be embedded into another hook that is used to create a component

JordanMartinez:
Aside from the above comment and #120, this issue is basically done.

milesfrain:
> I'm not sure whether the following recipes should be implemented based on their readme's content. Thoughts?

Yes, that definitely looks like duplication.
It would be good to note this for readers attempting to find equivalent recipes to the examples. Not sure where to put this note. Maybe add something like this to the [examples readme](https://github.com/thomashoneyman/purescript-halogen-hooks/blob/master/examples/README.md):
[...]

PR #162, opened 2020-07-10
Add AceHalogenHooks

JordanMartinez:
Addresses part of #138 

Until `ace` is included in the package set, this cannot be merged.


recipes/AceEditorHalogenHooks/src/Main.purs
```diff
@@ -0,0 +1,100 @@
...
+  state /\ stateIdx <- Hooks.useState Nothing
+  Hooks.useLifecycleEffect do
+    Hooks.getHTMLElementRef (H.RefLabel "ace") >>= traverse_ \element -> do
+      editor <- liftEffect $ Ace.editNode element Ace.ace
+      session <- liftEffect $ Editor.getSession editor
+      Hooks.put stateIdx $ Just editor
+      void $ Hooks.subscribe $ ES.effectEventSource \emitter -> do
+        Session.onChange session \_ -> ES.emit emitter do
+          Hooks.get stateIdx >>= traverse_ \editor' -> do
+            text <- liftEffect (Editor.getValue editor')
```
milesfrain, line 81: Among `Editor`, `Session`, and `Document`, what should be the main interface?
In this example, it's:
* `Editor` saved as `state`
* `Session` `onChange` detected
* `Editor` `getValue` and `setValue`

It's also possible to do all of the above via `Document`, which I think is more straightforward if we're just looking for changes to text https://ace.c9.io/#nav=api

recipes/AceEditorHalogenHooks/src/Main.purs
```diff
@@ -0,0 +1,100 @@
...
+data AceQuery a = ChangeText String a
+
+data AceOutput = TextChanged String
+
+aceComponent
+  :: forall unusedInput anyMonad
+   . MonadAff anyMonad
+  => H.Component HH.HTML AceQuery unusedInput AceOutput anyMonad
+aceComponent = Hooks.component \rec _ -> Hooks.do
+  state /\ stateIdx <- Hooks.useState Nothing
```
milesfrain, line 72: It's nice to use annotations when the type is not obvious from the initial value.
For example:
```purs
Hooks.useState (Nothing :: Maybe Editor)
```
Although I'm thinking `Maybe Document` would be better (as mentioned earlier).
Also, should `state` be renamed to something more specific? Downside is that requires some de-shadowing with duplicated names.

recipes/AceEditorHalogenHooks/src/Main.purs
```diff
@@ -0,0 +1,100 @@
...
+  Hooks.useLifecycleEffect do
+    Hooks.getHTMLElementRef (H.RefLabel "ace") >>= traverse_ \element -> do
```
milesfrain, line 74: Since `(H.RefLabel "ace")` is used twice, what about a common `aceRefLabel` to avoid this "magic number / string"?

milesfrain:
Rather than tracking `web/ace-1.1.3.js` you can add `"ace-builds": "^1.4.11",` to `package.json`.

recipes/AceEditorHalogenHooks/src/Main.purs
JordanMartinez: Good point. I've resolved that in latest commits.

JordanMartinez:
> Rather than tracking web/ace-1.1.3.js you can add "ace-builds": "^1.4.11", to package.json.

Yeah. My goal this morning was just to get `make <name>-web` to compile for this recipe and copying that over was the easiest.
This has now been resolved in latest commits.

recipes/AceEditorHalogenHooks/src/Main.purs
JordanMartinez: I'm not sure. I was just porting over the example, not trying to make this version of the example be more optimized than the original.

JordanMartinez:
Pending purescript/package-sets#656

JordanMartinez:
The package set PR was merged. Just waiting for a new release now.

recipes/AceEditorHalogenHooks/src/Main.purs
JordanMartinez: I did use clarify the `Nothing` type [in this commit](https://github.com/JordanMartinez/purescript-cookbook/pull/162/commits/74295e720a7557faa0a5405581cd66d7562023b5#diff-8a44b4282b62e6e07657f75e3dba6ef7R76)

Not sure about the `state` renaming...

recipes/AceEditorHalogenHooks/src/Main.purs
JordanMartinez: I'd be up for looking at this again in #184

PR #163, opened 2020-07-10
Use useAff's result instead of synchronizing with separate state

spicydonuts:
This is a great resource, thanks for starting it!

I think for simple "fetch and display" tasks it's better to use `useAff`'s result directly rather than synchronizing it with a separate state. Do you prefer the separate state type? (this is mostly for my curiosity, with the PR just providing a diff to look at -- feel free to ignore/close).

### DIFF
--- a/recipes/BookReactHooks/src/Main.purs
+++ b/recipes/BookReactHooks/src/Main.purs
@@ -21,7 +21,6 @@ import Web.HTML.Window (document)
 
 data TextState
   = Failure
-  | Loading
   | Success String
 
 main :: Effect Unit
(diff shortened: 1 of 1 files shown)


ptrfrncsmrph:
I'm absolutely open to (and very much appreciate) this type of feedback, since I'm learning PureScript by contributing here.

> Do you prefer the separate state type?

I think the only reason I had the separate state type was to mirror the Elm example which had
```elm
type Model
  = Failure
  | Loading
  | Success String
```
But I think it'd be more valuable to just have idiomatic PureScript/`react-basic-hooks` in this example.

JordanMartinez:
Looks cleaner to me. I'll merge.

PR #166, opened 2020-07-12
Implementation of the HelloWorld receipt using Concur

gcsolaroli:
I have ported the simplest receipt to Concur; if the result is suitable for inclusion in the cookbook, I will keep on porting the other receipts.

### DIFF
--- a/recipes/HelloConcur/README.md
+++ b/recipes/HelloConcur/README.md
@@ -1,6 +1,6 @@
 # HelloConcur
 
-A Concur port of the ["HTML - Hello" Elm Example](https://elm-lang.org/examples).
+A Concur port of the ["HTML - Hello" Elm Example](https://elm-lang.org/examples/hello).
 
 ## Expected Behavior:
 
--- a/recipes/HelloConcur/README.md
+++ b/recipes/HelloConcur/README.md
@@ -1,0 +1,9 @@
+# HelloConcur
+
+A Concur port of the ["HTML - Hello" Elm Example](https://elm-lang.org/examples).
+
+## Expected Behavior:
+
+### Browser
+
+The browser will display "Hello!" as an html element.
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ Running a web-compatible recipe:
 |   | :heavy_check_mark: | [FormsReactHooks](recipes/FormsReactHooks) | A React port of the ["User Interface - Forms" Elm Example](https://elm-lang.org/examples/forms). |
 |   | :heavy_check_mark: | [GroceriesHalogenHooks](recipes/GroceriesHalogenHooks) | A Halogen port of the ["HTML - Groceries" Elm Example](https://elm-lang.org/examples). |
 |   | :heavy_check_mark: | [GroceriesReactHooks](recipes/GroceriesReactHooks) | A React port of the ["HTML - Groceries" Elm Example](https://elm-lang.org/examples). |
+|   | :heavy_check_mark: | [HelloConcur](recipes/HelloConcur) | A Concur port of the ["HTML - Hello" Elm Example](https://elm-lang.org/examples/hello). |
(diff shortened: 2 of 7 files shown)


recipes/HelloConcur/README.md
```diff
@@ -0,0 +1,9 @@
+# HelloConcur
+
+A Concur port of the ["HTML - Hello" Elm Example](https://elm-lang.org/examples).
+
+## Expected Behavior:
+
+### Browser
+
+The browser will display "Hello!" as an html element.
```
milesfrain, line 9: Are there any `npm` dependencies for Concur with react background?
For example, here are the dependencies listed for `HelloReactHooks` https://github.com/JordanMartinez/purescript-cookbook/blob/master/recipes/HelloReactHooks/README.md

milesfrain:
The CI failure can be fixed by running `make readme` to update the main readme table.

Should we specify the Concur backend by naming this `HelloConcurReact`?

What are all the available backends at the moment? Was there discussion about adding alternate Concur backends? @ajnsit for clarification.

recipes/HelloConcur/README.md
```diff
@@ -0,0 +1,9 @@
+# HelloConcur
+
+A Concur port of the ["HTML - Hello" Elm Example](https://elm-lang.org/examples).
```
milesfrain, line 3:
```suggestion
A Concur port of the ["HTML - Hello" Elm Example](https://elm-lang.org/examples/hello).
```

gcsolaroli: For consistency, this same change should be applied also to the README file in the HelloHalogenHooks folder.

recipes/HelloConcur/README.md
gcsolaroli: I missed that listing of the dependencies; I have just updated spago.dhall adding `concur-react`.

JordanMartinez:
In #123, it was my understanding that we would use `*Concur` suffix for Concur-based recipes.

recipes/HelloConcur/README.md
milesfrain: I think it's helpful to list the required npm dependencies in case a reader wants to reuse recipe code in their project.

For example, if the code for this recipe is copied into a fresh `spago init` project, it won't work. The reader must also install the `react` and `react-dom` dependencies.

The npm dependencies for all recipes are shared in this common file:
https://github.com/JordanMartinez/purescript-cookbook/blob/master/package.json

It would be better to track these per recipe, but we haven't found a great way to do this yet, so we're just noting these in the readme.

ajnsit: `concur-react` has `react` and `react-dom` as npm dependencies. The upcoming `concur-vdom` will have no npm dependencies.

ajnsit:
> What are all the available backends at the moment? Was there discussion about adding alternate Concur backends? @ajnsit for clarification.

`Concur-React` is the currently available backend for Purescript. There's a work in progress `Concur-Vdom` backend that uses `halogen-vdom` which should be available soon.

milesfrain:
So I'm fine with `*Concur` suffix for now, and then we can convert to `vdom` backend later without renaming.
This is exciting, since no `npm` dependencies means we'll then be able to generate TryPureScript links to these. (currently working on automating that for our recipe table with a TPS rewrite).

ajnsit:
Actually, Concur already works with TryPurescript. For example, here's an svg dragon curve example - https://try.purescript.org/?gist=c3637f0be0e8857c4c118ae6a35c3663

JordanMartinez:
Would Concur work on both React and Halogen-VDOM? If so, would we want to have examples for both?

ajnsit:
Yes, both the backends will be available, so using `ConcurReact` and `ConcurVdom` as prefixes is a good idea. There would be other backends in the future which can follow the same scheme. I've been thinking of adding an HTML canvas backend, which could be `ConcurCanvas`.

JordanMartinez:
Ok, could we update the list of targets in a separate PR to clarify that?

PR #173, opened 2020-07-13
Add HeterogenousArrayLog

### LINKED ISSUE #149: How to create a heterogeneous array in Purescript
**Recipe Name**

```
HeterogenousArray
```

**Recipe Description**

Briefly explain that the type system explicitly prohibits naive heterogeneous collections and show how one can achieve the result when necessary using `purescript-variant`

JordanMartinez:
Fixes #149

### DIFF
--- a/recipes/HeterogenousArrayLog/README.md
+++ b/recipes/HeterogenousArrayLog/README.md
@@ -8,6 +8,17 @@ PureScript arrays must store values that have the same type. Attempting to store
 
 Prints the following:
 ```
+[Str(a String value),Integer(4),Bool(true),Bool(false),Num(82.4)]
+------------------
+["a String value","4","true","false","82.4"]
+------------------
+String value: a String value
+Int value: 4
+Boolean value: true
+Boolean value: false
+Number value: 82.4
+------------------
+------------------
 [(inj @"typeLevelString" "a String value"),(inj @"int" 4),(inj @"boolean" true),(inj @"boolean" false),(inj @"this type level string must match the label of the row" 82.4)]
 ------------------
 ["(inj @\"typeLevelString\" \"a String value\")","(inj @\"int\" 4)","(inj @\"boolean\" true)","(inj @\"boolean\" false)","(inj @\"this type level string must match the label of the row\" 82.4)"]
--- a/recipes/HeterogenousArrayLog/README.md
+++ b/recipes/HeterogenousArrayLog/README.md
@@ -2,6 +2,10 @@
 
 This recipe shows how to create, display, and pattern match on a heterogenous array.
 
+PureScript arrays must store values that have the same type. Attempting to store values of different types will result in a compiler error.
+
+To get around this limitation while still ensuring type safety, we can use `Variant`.
+
 ## Expected Behavior:
 
 Prints the following:
--- a/recipes/HeterogenousArrayLog/README.md
+++ b/recipes/HeterogenousArrayLog/README.md
@@ -2,9 +2,7 @@
 
 This recipe shows how to create, display, and pattern match on a heterogenous array.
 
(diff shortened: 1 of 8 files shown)


JordanMartinez:
@milesfrain @afcondon Thoughts?

recipes/HeterogenousArrayLog/src/Main.purs
```diff
@@ -0,0 +1,87 @@
...
+separateWithDashes = log "------------------"
+
+main :: Effect Unit
+main = do
+  {-
+    PureScript arrays must store values that have the same type. Attempting
+    to store values of different types will result in a compiler error.
+
+    To get around this limitation while still ensuring type safety, we can use
+    `Variant`.
```
milesfrain, line 21: I think this note along with some more explanation could go in the readme. Content after line three won't be added to the table, so we can feel free to be more verbose in there.
I can see this topic being a bit unapproachable for beginners if the cookbook is their first exposure.

afcondon:
> @milesfrain @afcondon Thoughts?

One way to make this approachable to the beginner is to clearly position it as a somewhat exceptional situation. You really don't want someone who knows very little of purely functional languages to think that you need all this type level boilerplate all the time. But at the same time, you don't want someone to conclude that there's no "escape hatch" should it be necessary.

So i definitely agree with @milesfrain that the README is crucial.

One thing to emphasize is the use of Sum types to create simple heterogeneous collections (maybe there's even another recipe in that, i don't know) and then motivate Variant by showing that they're a kind of Sum type but more flexible again. And that can lead to specifically why one might be using them for logging, i guess? 

that's my $0.02 from a quick look over the code and comments first thing this morning

afcondon:
Oh, there's also this that i just found - i don't know if that's a stepping stone to the logging example or a distraction but i thought i'd share it just FYI: https://github.com/justinwoo/purescript-kishimen

recipes/HeterogenousArrayLog/src/Main.purs
JordanMartinez: Good point.

Do you think we should also indicate difficulty in the recipe?

JordanMartinez:
I've incorporated your feedback in latest commits.

milesfrain:
I feel like this recipe could be simplified and still provide the same kernel of information about heterogeneous arrays. Let me know if there's any core content you find missing from my rewrite.

recipes/HeterogenousArrayLog/README.md
```diff
@@ -0,0 +1,16 @@
+# HeterogenousArrayLog
+
+This recipe demonstrates how to create a heterogenous array and process its elements.
+
+PureScript arrays are _homogeneous_, meaning that all values must have the same type. If you want to store values with different types in the same array (i.e. a _heterogeneous_ array), you should wrap all the types you wish to store in either a [sum type](https://github.com/purescript/documentation/blob/master/language/Types.md#tagged-unions) or a [Variant](https://pursuit.purescript.org/packages/purescript-variant). This recipe demonstrates both strategies.
+
```
milesfrain, line 6: A note about pros and cons of either approach may be helpful. I'm not aware of any besides personal preference for style.

JordanMartinez: Variant is extensible, so it comes with the pros/cons of that. However, anything row-related will likely incur a performance cost in some situations due to type class dictionaries.

JordanMartinez:
No, I think your rewrite was good. Do you think we should reduce the number of types down to just two? I don't think we need to show 3-4 types in a single Array.

milesfrain:
I don't think the extra few types adds too much noise, but no string opinions on that.

JordanMartinez:
Ok. I think I'll merge this then.

PR #187, opened 2020-07-17
Memoize fibonacci

### LINKED ISSUE #182: MemoizeFibonacci
**Recipe Name**

```
MemoizeFibonacci
```

**Recipe Description**

Describes how to use `memoize`, which I found pretty tricky to get right.

Here's the implementation: https://try.purescript.org/?gist=a1eca7355ed7e3d6c1198956dde8c1fe
Open dev console to view logs.

Will use this recipe to test workflow for #181

milesfrain:
Fixes #182

### DIFF
--- a/recipes/MemoizeFibonacci/src/Main.purs
+++ b/recipes/MemoizeFibonacci/src/Main.purs
@@ -7,6 +7,12 @@ import Debug.Trace (spy)
 import Effect (Effect)
 import Effect.Class.Console (log)
 
+main :: Effect Unit
+main = do
+  log $ i "basic fib result: " $ fib 7
+  log $ i "fibFast result: " $ fibFast 7
+  log $ i "fibSlow result: " $ fibSlow 7
+
 -- Basic fibonacci implementation
 fib :: Int -> Int
 fib 0 = 0
@@ -44,10 +50,3 @@ type-check. Feel free to experiment with these yourself:
   * `fibMemo = memoize fibFast`
 * Move to a `let` or `where` clause inside of `fibFast`.
 -}
-
-
-main :: Effect Unit
-main = do
-  log $ i "basic fib result: " $ fib 7
-  log $ i "fibFast result: " $ fibFast 7
-  log $ i "fibSlow result: " $ fibSlow 7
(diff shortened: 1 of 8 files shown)


recipes/MemoizeFibonacci/src/Main.purs
```diff
@@ -0,0 +1,53 @@
...
+{-
+Note that the following adjustments to `fibMemo` eliminate the
+performance benefits of `memoize`, even though they may still
+type-check. Feel free to experiment with these yourself:
+
+* Rewrite in non-thunked or point-free form:
+  * `fibMemo n = memoize fibFast n`
+  * `fibMemo = memoize fibFast`
+* Move to a `let` or `where` clause inside of `fibFast`.
+-}
```
JordanMartinez, line 46: If I'm understanding you correctly, adjusting `fibMemo` in the above 3 ways will remove the performance benefit of memoization. 
Could we convert those examples to code, add them as statements to `main` but leave those statements commented out with a header above explaining why?
For example:
```purescript
main :: Effect Unit
main = do
  log $ i "basic fib result" $ fib 7
  ...
  log $ i "fibSlow result: " $ fibSlow 7

  -- These examples type check and seem to implement
  -- to incur the performance benefits of memoization. 
  -- However, they do not. Uncomment them to see for yourself.
  -- log $ i "fibBroken1 result: " $ fibBroken1 7
  -- ..
```
I think I'd also prefer that `main` be immediately below the import statements, so that one can see what will run and then see how they are defined.

Thoughts?
```

JordanMartinez: ping @milesfrain

milesfrain: Yeah. These are good suggestions. I'll circle back to this task in about a week. Trying to wrap-up the try-purescript enhancements first.

JordanMartinez: In that case, I'll push changes to this PR and merge. If you have feedback on what I've done, feel free to open an issue.

Stay focused on the try-purescript stuff as that's more important.

PR #189, opened 2020-07-18
Fix broken CatGifsReactHooks recipe

### LINKED ISSUE #183: Fix `CatGifsReactHooks`
**Recipe Name**

```
CatGifsReactHooks
```

**Reason for Breakage / Steps to Fix**

Updated package set from `psc-0.13.6-202005xx` to `psc-0.13.6-20200716`. Seems there was a breaking change related to `AffReducer` in React. See [this comment and its PR](https://github.com/JordanMartinez/purescript-cookbook/pull/178#issuecomment-657931737) for more info.

ptrfrncsmrph:
Resolves #183


recipes/CatGifsReactHooks/src/Main.purs
```diff
@@ -55,33 +39,34 @@ main = do
 mkCatGifs :: Component {}
 mkCatGifs = do
   component "CatGifs" \_ -> React.do
-    (gifState /\ dispatch) <- useAffReducer Loading reducer
```
ptrfrncsmrph, line 58: Removed the use of `useAffReducer` here. The version of `react-basic-hooks` that's in the package set we're using is `v6.0.0`, and it seems that version doesn't export the `mkAffReducer` that's needed to construct an `AffReducer` 🙈 . This lead me to discover `useResetToken`, which the author seems to recommend over `useAffReducer` anyways, and it ends up being less code and a bit simpler 👍 .

recipes/CatGifsReactHooks/src/Main.purs
```diff
@@ -55,33 +39,34 @@ main = do
...
       ResponseFormat.json
       "https://api.giphy.com/v1/gifs/random?api_key=dc6zaTOxFJmzC&tag=cat"
   pure do
-    { body } <- lmap printError response
-    body' <- decodeJson body
-    data' <- body' .: "data"
-    data' .: "image_url"
+    -- Using `hush` to ignore the possible error messages
+    { body } <- hush response
+    hush (decodeJson body >>= (_ .: "data") >>= (_ .: "image_url"))
```
ptrfrncsmrph, line 72: There was also a breaking change with `decodeJson` which now returns an `Either JsonDecodeError` instead of `Either String`. Instead of trying to unify the error types, I opted to just convert it to `Maybe`.

milesfrain:
Nice. Very efficient code.
Hopefully this isn't too confusingly concise for beginners:
```ps
toGifState :: Maybe (Maybe String) -> GifState
toGifState = maybe Loading (maybe Failure Success)
```
But they're free to ask clarifying questions, so LGTM.

ps. Even though Jordan gave the go-ahead for me to merge PRs, I'd like to give some time for collecting feedback.

ptrfrncsmrph:
Sounds good 👍 . I'm learning PS by contributing here, so I value feedback!

JordanMartinez:
Thanks for fixing this!

Yeah, the `toGifState` is a bit different, but the tradeoffs aren't really worth it either. I'm not sure when we should err on being more readable vs being more idiomatic/abbreviated.

recipes/CatGifsReactHooks/src/Main.purs
JordanMartinez: Thanks for clarifying this!

Issue #157, opened 2020-07-08
Elm Examples with React


milesfrain:
Extends #86

@ptrfrncsmrph is working on this https://github.com/JordanMartinez/purescript-cookbook/issues/86#issuecomment-653722944

- [x] HelloReactHooks
- [x] GroceriesReactHooks
- [x] ShapesReactHooks
- [x] ButtonsReactHooks
- [x] TextFieldsReactHooks
- [ ] FormsReactHooks
- [x] NumbersReactHooks
- [x] CardsReactHooks
- [x] PositionsReactHooks
- [x] BookReactHooks
- [x] CatGifsReactHooks
- [ ] TimeReactHooks
- [ ] ClockReactHooks
- [ ] UploadReactHooks
- [ ] DragAndDropReactHooks
- [ ] ImagePreviewsReactHooks

ptrfrncsmrph:
Just to give a timeline on the ones I haven’t got to yet: I’ll plan on getting Shapes, Forms, Time and Upload within a week from now, and then I’ll claim more at that point if there are any still remaining.

JordanMartinez:
Remember to use the `ReactHooks` suffix on the recipe name.

PR #190, opened 2020-07-19
Add ClockReactHooks recipe

ptrfrncsmrph:
Partially addresses #157

### DIFF
--- a/recipes/ClockReactHooks/src/Main.purs
+++ b/recipes/ClockReactHooks/src/Main.purs
@@ -27,13 +27,13 @@ main = do
   case container of
     Nothing -> throw "Root element not found."
     Just c -> do
-      time <- mkTime
-      render (time {}) c
+      clock <- mkClock
+      render (clock {}) c
 
-mkTime :: Component {}
-mkTime = do
+mkClock :: Component {}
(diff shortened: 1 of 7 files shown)


recipes/ClockReactHooks/src/Main.purs
```diff
@@ -0,0 +1,89 @@
...
+              , hand 6 60.0 (hours / 12.0)
+              , hand 6 90.0 (minutes / 60.0)
+              , hand 3 90.0 (seconds / 60.0)
+              ]
+          }
+
+hand :: Int -> Number -> Number -> JSX
+hand width length turns =
+  let
+    t = 2.0 * pi * (turns - 0.25)
```
milesfrain, line 54:
```suggestion
    t = tau * (turns - 0.25)
```
https://pursuit.purescript.org/packages/purescript-math/docs/Math#v:tau

ptrfrncsmrph: Good call! I had blindly copied that from the Elm example verbatim, but looks like Elm just has `pi` and not `tau` in its standard library.

JordanMartinez:
Thanks!

PR #192, opened 2020-07-23
Implement DateTimeBasicsLog

### LINKED ISSUE #94: DateTime basics
**Description of Problem**

Show how to use `purescript-datetime` in various ways:

* construct a Time value and add a Duration to it
* construct two DateTime values and find the difference between them

```
Unique Recipe Name: DateTimeBasics
```

JordanMartinez:
Fixes #94

### DIFF
--- a/recipes/DateTimeBasicsLog/src/Main.purs
+++ b/recipes/DateTimeBasicsLog/src/Main.purs
@@ -1,0 +1,93 @@
+module DateTimeBasicsLog.Main where
+
+import Prelude
+
+import Data.Date (Date, Month(..), canonicalDate, exactDate)
+import Data.DateTime (DateTime(..), adjust, diff)
+import Data.Enum (toEnum)
+import Data.Maybe (Maybe, fromJust)
+import Data.Time (Hour, Minute, Time(..))
+import Data.Time.Duration as D
+import Effect (Effect)
+import Effect.Console (log)
+import Partial.Unsafe (unsafePartial)
+
+main :: Effect Unit
+main = do
+  log "To create a specific part of a `Time` value, you must use \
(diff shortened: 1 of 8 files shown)


osfameron:
This is great, really useful documentation for starting with these modules!

The definition of `mkTime :: Int -> Int -> Int -> Int -> Maybe Time` and co look so generally useful, that would it be worth submitting them as PRs to the library itself?

JordanMartinez:
> The definition of mkTime :: Int -> Int -> Int -> Int -> Maybe Time and co look so generally useful, that would it be worth submitting them as PRs to the library itself?

Probably. `Int` arguments don't clarify what their corresponding values are. An example in the docs above such a function would probably clear that up.

recipes/DateTimeBasicsLog/src/Main.purs
```diff
@@ -0,0 +1,93 @@
...
+import Data.Maybe (Maybe, fromJust)
+import Data.Time (Hour, Minute, Time(..))
+import Data.Time.Duration as D
+import Effect (Effect)
+import Effect.Console (log)
+import Partial.Unsafe (unsafePartial)
+
+main :: Effect Unit
+main = do
+  log "To create a specific part of a `Time` value, you must use \
```
thomashoneyman, line 17: In the places where we need to long strings like this in our code at work we'll usually use `"""` as in:

```purs
"""
To create a specific part of a `Time` value, you must use the `toEnum`
function. This ensures the value you provide is within the correct bounds.
"""
```

which saves having to add all the `\`. Not necessary, but thought I'd share it in case it helps with writing some of these long strings.

JordanMartinez: That's a good point. I'd prefer that style.

PR #195, opened 2020-08-03
Add (try) links to Try PureScript demo for web recipes

milesfrain:
See table here: https://github.com/milesfrain/purescript-cookbook/blob/try-links/README.md#recipes

We can include some additional refinements to control whether to generate these links depending on whether the recipe is actually compatible with the try purescript environment.

Related to: https://github.com/purescript/trypurescript/pull/187


milesfrain:
Looks like the react recipes could use some troubleshooting:
https://try.ps.ai/?github=JordanMartinez/purescript-cookbook/master/recipes/CardsReactHooks/src/Main.purs
See error in dev console.

The halogen versions work fine https://try.ps.ai/?github=JordanMartinez/purescript-cookbook/master/recipes/CardsHalogenHooks/src/Main.purs

JordanMartinez:
Very nice!

I tried the above link and got this error:
```
bad result. Likely communication issue with compiler: POST compile response failed to decode: There was a problem making the request: AJAX request failed: POST https://compile.purescript.org/compile
```

milesfrain:
I can't reproduce that compiler error. Could you try again, and if the issue persists, could we setup a screen share meeting for troubleshooting?

JordanMartinez:
I can in fifteen minutes.

JordanMartinez:
Nevermind. Turns out a browser extension (Privacy Badger in this case) was blocking the request.

thomashoneyman:
> Nevermind. Turns out a browser extension (Privacy Badger in this case) was blocking the request.

That may be worth adding to the error message (a privacy extension may be blocking the request), as these extensions are increasingly common and strict.

PR #200, opened 2020-08-08
Note how ImagePreviews is different than DragAndDrop

milesfrain:


milesfrain:
Forgot to run the readme script, but I'll do that after #206, since they touch the same line and will conflict.

PR #210, opened 2020-08-14
Add HelloJs and GroceriesJs

milesfrain:
Demos:

https://try.ps.ai/?github=milesfrain/purescript-cookbook/hello-groceries-js/recipes/HelloJs/src/Main.purs

https://try.ps.ai/?github=milesfrain/purescript-cookbook/hello-groceries-js/recipes/GroceriesJs/src/Main.purs

PR #211, opened 2020-08-14
Reduce groceries verbosity

milesfrain:
Intended to be merged after #210 

For larger apps, it's good to prefix the HTML tags, but in this case, I don't want newcomers comparing these examples to Elm and concluding that PS syntax is terrible.

### DIFF
--- a/recipes/GroceriesHalogenHooks/src/Main.purs
+++ b/recipes/GroceriesHalogenHooks/src/Main.purs
@@ -4,10 +4,10 @@ import Prelude
 import Data.Maybe (Maybe(..))
 import Effect (Effect)
 import Halogen as H
-import Halogen.HTML as HH
 import Halogen.Aff as HA
-import Halogen.VDom.Driver (runUI)
+import Halogen.HTML (HTML, div_, h1_, li_, text, ul_)
 import Halogen.Hooks as Hooks
+import Halogen.VDom.Driver (runUI)
 
 main :: Effect Unit
 main =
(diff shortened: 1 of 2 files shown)


JordanMartinez:
While I get your point, it's also conventional to use the `HH` prefix. So, I'm hesitant about merging this.

milesfrain:
Lets leave this open for a few days to ponder.

Comparison links for reference:
[Existing](https://try.ps.ai/?github=JordanMartinez/purescript-cookbook/master/recipes/GroceriesHalogenHooks/src/Main.purs) Halogen - [Proposed](https://try.ps.ai/?github=milesfrain/purescript-cookbook/groceries-cleanup/recipes/GroceriesHalogenHooks/src/Main.purs) Halogen - [Elm](https://elm-lang.org/examples/groceries)

I think it's the red text color in TPS that's making these prefixes really distracting.


I've been treating the `HH.` prefix as a looser convention in my own projects. Not sure if deviating from convention occasionally in these recipes is confusing for readers, or a nice reminder that there's flexibility in how to manage imports.
I have no plans for removing the `HH.` prefix from other recipes, since those don't have the same level of duplication.

and-pete:
> I've been treating the `HH.` prefix as a looser convention in my own projects. Not sure if deviating from convention occasionally in these recipes is confusing for readers, or a nice reminder that there's flexibility in how to manage imports.
> I have no plans for removing the `HH.` prefix from other recipes, since those don't have the same level of duplication.

I think it would be great to have a section in `purescript-faq` about the what and why behind the conventions for imports in PureScript (...and potentially for Halogen also).

afcondon:
You could go with prefix-less version but put the existing code in a block comment with a little informative banner on it, saying that in practice any large project will probably _choose_ to use prefixes as shown but that they are not in any way required by the library or the language.

Could be like those little inset boxes in some text books that cover "Advanced Topic" or "Hot Tip" kind of thing but which you know you can skip the first time reading it.

milesfrain:
> I think it would be great to have a section in `purescript-faq` about the what and why behind the conventions for imports in PureScript

[The docs](https://github.com/purescript/documentation/blob/master/language/Modules.md#qualified-imports) explain this pretty well. I think this would be more discoverable if we model our docs on the strategy used by ReasonML, where there's a [quick overview](https://reasonml.github.io/docs/en/overview), and then links to additional details for each topic.

> put the existing code in a block comment

This seems like a good compromise.

JordanMartinez:
>> put the existing code in a block comment
> This seems like a good compromise.

I'd be good with that!

JordanMartinez:
@milesfrain I clarified why the HH prefix was not used in latest commit. I put it at the bottom rather than above the function or within it, so that the code continues to appear very readable. Otherwise, I thought the code block might distract the reader's attention.

If you're good with these changes, let's get this merged.

PR #212, opened 2020-08-14
Add SignalRenderJs

milesfrain:
Demo:
https://try.ps.ai/?github=milesfrain/purescript-cookbook/signals-render-js/recipes/SignalRenderJs/src/Main.purs

Wanted to add some simple examples of using the [signal](https://github.com/bodil/purescript-signal) library. This example may even be too complex. I may eventually move some of this to the Canvas Graphics chapter of the book with a game-writing tutorial, but figured it would be good to stash some examples in the cookbook for the meantime.

We could also host some games here too, since it's convenient to be able to open them directly in TPS.

I'd also like to show the equivalent of each signal recipe using [behaviors](https://github.com/paf31/purescript-behaviors), but that's trickier for me to figure out.


JordanMartinez:
Looks like you forgot to regenerate the readme.

PR #213, opened 2020-08-15
Add SignalSnakeJs

milesfrain:
Demo:
https://try.ps.ai/?github=milesfrain/purescript-cookbook/signal-snake-js/recipes/SignalSnakeJs/src/Main.purs

If this example isn't too confusing for beginners, it could replace #212.


milesfrain:
Keeping this as a draft to give @holdenlee a chance to share citation preferences.
(Discussed in https://github.com/holdenlee/purescript-games/issues/4)

recipes/SignalSnakeJs/src/Main.purs
```diff
@@ -0,0 +1,369 @@
...
+-- is used here to provide a signal value at time = 0.
+mapKey :: Direction -> Signal Boolean -> Signal Direction
+mapKey dir sig = filterMap (fromBool dir) initialDirection sig
+
+-- Helper function for mapKey's filterMap.
+-- Convert's a Boolean to a Maybe using a default value.
+fromBool :: forall a. a -> Boolean -> Maybe a
+fromBool x b
+  | b = Just x
+  | otherwise = Nothing
```
milesfrain, line 275: I suspected something like this would already exist, but I couldn't find anything on Pursuit.

JordanMartinez:
I don't believe we should merge this until a license for that work is issued. I can't find any on the repo.

That's the only blocker in my mind.

holdenlee:
I give permission to use the example in the PureScript Cookbook.

afcondon:
One style suggestion (and it could well be that you disagree with it, but i'll just put it here for consideration) is that you could use a case instead of the nested if-then-else, ie

```    
    case willDie, willEat of
         false  , false -> m { snake = moveSnakeNoGrow nextPoint m.snake }
         true   , _     -> m { snake = initialSnake, direction = initialDirection }
         false  , true  -> 
          let 
            biggerSnake = moveSnakeAndGrow nextPoint m.snake
            food /\ genState = runGen (availableRandomPoint biggerSnake) m.genState
          in
            m
              { snake = biggerSnake
              , food = food
              , genState = genState
              }
```

milesfrain:
> use a case instead of the nested if-then-else

Good suggestion. Now if only [purty would let me keep those alignment spaces](https://gitlab.com/joneshf/purty/-/issues/200)...

afcondon:
Yes, i love the existence of, and many ideas in, Purty, but to me compromising readability of the actual code in order to maximize readability of commits is putting the cart before the horse.

Very possibly not worth doing if formatting messes it up, if-then-else isn't the worst thing in the world.

recipes/SignalSnakeJs/src/Main.purs
paluh: I think that Nate showed this on the slack recently:

```purescript
import Control.MonadZero (guard)

fromBool x b = guard b $> x
```

JordanMartinez:
> I give permission to use the example in the PureScript Cookbook.

@holdenlee Thanks, but under what terms? Per our [guidelines here](https://github.com/JordanMartinez/purescript-cookbook/blob/master/CONTRIBUTING.md#all-recipes-are-licensed-under-this-repos-license-mit), all recipes are licensed under MIT. Are you ok with that?

holdenlee:
@JordanMartinez Yes, under the MIT license.

PR #224, opened 2020-08-23
Add SignalTrisJs

milesfrain:
Demo: https://try.purescript.org/?github=/JordanMartinez/purescript-cookbook/master/recipes/SignalTrisJs/src/Main.purs

@cheery 
This is a fun example that I'd like to include in the cookbook. I rewrote this to use the Signals library, which I think works better for games. Let us know if we have your permission to include this in the cookbook and whether you'd like to be credited any differently in the readme. Also feel free to critique the code before it's merged. More reviewers are ideal.

Here are the incremental changes of the rewrite: https://github.com/milesfrain/tris/pull/1


cheery:
I already approved this kind of use on my blog, with the license I pick just recently. Thank you for keeping me involved though. I appreciate that.

I find the code itself helpful because you've improved on it. I am also satisfied to the crediting. You are very much welcome to merge this commit.

I'll put an update to the page that the Tris can be found from this cookbook as well.

JordanMartinez:
Thanks all!

PR #226, opened 2020-08-25
Payload HTTP server backend recipe

### LINKED ISSUE #225: Payload HTTP Backend Example
**Recipe Name**

```
PlayloadHttpApiNode
```

Each recipe name must be unique. Use PascalCase.

**Recipe Description**

A simple API example for the strongly typed HTTP backend payload. Ideally, it would show an API that allows for get all, get by id and insert, at least.

I already implemented something using `AVar`, and I was wondering whether this is too complicated for a single recipe? The alternative would be using files instead (to keep state across calls), but then it would also complicates matters due to encoding/decoding.

Do you think it would be an useful recipe?

PS: I cannot use `StateT` because it's not currently possible yet (in payload). I added an issue, and am willing to update this when/if it gets resolved.

vladciobanu:
Fixes #225 

This currently overrides the packages.dhall additions for it to work. I created a PR on package-set to add it there https://github.com/purescript/package-sets/pull/675

Is it okay if this just stays open until then? If not, feel free to close and I can re-open.


recipes/PayloadHttpApiNode/README.md
```diff
@@ -0,0 +1,27 @@
+# PayloadHttpApiNode
+
+Implements a simple 'quote' API using the [payload](https://github.com/hoodunit/purescript-payload)
+HTTP backend.
```
JordanMartinez, line 4: Since this is on line 4, it won't be included when you run `make readme`. Remove the newline and let the text wrap.

JordanMartinez: Also, could you add something in its description that says this has similar design principles as Haskell's Servant?

recipes/PayloadHttpApiNode/nodeSupported.md
```diff
@@ -0,0 +1,2 @@
+This file just indicates that the node backend is supported.
```
JordanMartinez, line 1: This file should be named `nodeSupportedSkipCI.md`. Otherwise, CI will run this code and it will never terminate.

JordanMartinez:
> Is it okay if this just stays open until then?

Yeah, that's fine! It's what I would do, too, in this situation.

Looks good to me!

vladciobanu:
PR was merged in package-sets, just need to wait for a new tag to get published.

JordanMartinez:
Looks like payload is now included in the recent package set release. @vladciobanu could you update this PR to use the latest release?

JordanMartinez:
Looks like you [accidentally added a `SuperCircle` into the readme](https://travis-ci.com/github/JordanMartinez/purescript-cookbook/builds/183226337#L327) and that's what's stopping CI from building. Mind removing that?

vladciobanu:
I accidentally commited the modified package.json file; I'll revert and fix it.

JordanMartinez:
Thanks for all your work on this!

PR #228, opened 2020-08-26
Convert template suggestions to table and other contributing edits

milesfrain:
Follow-up to https://github.com/JordanMartinez/purescript-cookbook/issues/225#issuecomment-680469009

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,7 +31,7 @@ Follow these instructions for contributing new recipes. The Goal headers indicat
     - **Web** - Recipe is compatible with the web browser backend.
     - **Node** - Recipe is compatible with the Node.js backend.
     - **CLI** - Recipe does not terminate, and/or requires user interaction (e.g. a CLI app) when run in the Node.js environment. These recipes contain a special configuration flag to prevent our CI process from stalling on these recipes during testing.
-    - **Framework** - A web framework for recipes that run in the browser.
+    - **Framework** - Recipe uses a particular web framework.
     - **Template** - A recommended recipe to copy.
 
 1. Run the recipe creation script. Substitute `MyNewRecipe` with your recipe name:
@@ -42,11 +42,14 @@ Follow these instructions for contributing new recipes. The Goal headers indicat
       ```
       ./scripts/newRecipe.sh MyNewRecipe recipes/HelloLog
       ```
-    - **Tip:** Note the `make` commands listed after running this script. These will be helpful when developing your recipe.
+    - **Tip:** The `make` commands printed after running this script will be helpful when developing your recipe. Make a copy of these for reference.
     - **Tip:** If you made a typo, or decide you want to rename your recipe, you can simply re-run this script to create a new recipe from your misspelled one with the correction, then delete the misspelled recipe:
       ```
       ./scripts/newRecipe.sh MyNewRecipeOops HelloLog
+      
       # *** Do lots of development work in the misspelled recipe ***
+      
+      # Fix the typo
       ./scripts/newRecipe.sh MyNewRecipe MyNewRecipeOops
       rm -r recipes/MyNewRecipeOops
       ```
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -53,15 +81,15 @@ Follow these instructions for contributing new recipes. The Goal headers indicat
 
 #### Only use packages in the latest package set release
 
-In other words, we do not allow anyone to override or add packages in the `packages.dhall` file for the following reasons:
```
milesfrain, line 56: Converting this to [BLUF](https://en.wikipedia.org/wiki/BLUF_(communication)) style where the key info is mentioned first, with justification at the end.

I noticed Vlad having to read though a bunch of stuff to get to the important bit about how to proceed with the recipe PR.

CONTRIBUTING.md
```diff
@@ -9,24 +9,47 @@ Follow these instructions for contributing new recipes. The Goal headers indicat
...
+    | :heavy_check_mark: | | | Halogen | `HelloHalogenHooks` |
+    | :heavy_check_mark: | | | React | `HelloReactHooks` |
+
+    Notes on interpreting the table:
+    - An empty cell means "not possible".
+    - **Web** - Recipe is compatible with the web browser backend.
+    - **Node** - Recipe is compatible with the Node.js backend.
+    - **CLI** - Recipe does not terminate, and/or requires user interaction (e.g. a CLI app) when run in the Node.js environment. These recipes contain a special configuration flag to prevent our CI process from stalling on these recipes during testing.
+    - **Framework** - A web framework for recipes that run in the browser.
+    - **Template** - A recommended recipe to copy.
```
milesfrain, line 35: The core change of this PR

CONTRIBUTING.md
```diff
@@ -9,24 +9,47 @@ Follow these instructions for contributing new recipes. The Goal headers indicat
 #### Goal 1: Claim and Start Working on a Recipe Request
 
 1. Search currently [open "Recipe Request" issues](https://github.com/JordanMartinez/purescript-cookbook/issues?q=is%3Aissue+is%3Aopen+label%3Arecipe-request).
-1. If your recipe idea is unique, open a new ["Recipe Request" issue](https://github.com/JordanMartinez/purescript-cookbook/issues/new?assignees=&labels=recipe-request&template=recipe-request.md&title=)
+1. (optional) If your recipe idea is unique, open a new ["Recipe Request" issue](https://github.com/JordanMartinez/purescript-cookbook/issues/new?assignees=&labels=recipe-request&template=recipe-request.md&title=). This step aims to prevents two people from working on the same recipe.
```
milesfrain, line 12: We talked about this before, and I can split this out to another PR for further discussion.
I think making this mandatory adds unnecessary overhead. Proposing we let this be optional until recipe collisions become a problem.

JordanMartinez: I'm good with this change. Name collisions haven't been an issue yet.

CONTRIBUTING.md
```diff
@@ -9,24 +9,50 @@ Follow these instructions for contributing new recipes. The Goal headers indicat
...
+
 1. Run the recipe creation script. Substitute `MyNewRecipe` with your recipe name:
     ```
     ./scripts/newRecipe.sh MyNewRecipe HelloLog
     ```
+    - **Tip:** You may point to the recipe path to take advantage of shell autocompletion
+      ```
+      ./scripts/newRecipe.sh MyNewRecipe recipes/HelloLog
+      ```
+    - **Tip:** The `make` commands printed after running this script will be helpful when developing your recipe. Make a copy of these for reference.
```
JordanMartinez, line 45: Perhaps we should add a command like `make help` that prints out these commands? If a person forgets to make note of them, they can just see it again by running that. The output would replace all instances of a recipe name with `<recipe-name>`

CONTRIBUTING.md
JordanMartinez: LGTM

CONTRIBUTING.md
JordanMartinez: This makes sense.

JordanMartinez:
I think the only thing we could do in addition to this PR is to add a `make help` command, which will print out the list of commands one can run after they create their new recipe. So, whether they remember to take note of them or not, they can still quickly review them again without having to leave the terminal.

CONTRIBUTING.md
milesfrain: Created #231 to track

milesfrain:
> add a `make help` command

Going to tackle that in #231

JordanMartinez:
Thanks!

Issue #232, opened 2020-08-29
Snake using AffGame


3ddyy:
**Recipe Name**

```
AffGameSnakeJs
```

**Recipe Description**

This recipe shows how to make a snake game using `purescript-game`

https://github.com/3ddyy/purescript-game/issues/3

I'll try to get this working today. `purescript-game` and two of its dependencies (all by me) are not yet in the package set though. Will get to that as well.

PR #233, opened 2020-08-29
Add AffGameSnakeJs

3ddyy:
Currently, `purescript-game` is unreleased and not in the package set. When i've written docs, released it, and gotten it in the package set, i will remove it from the additions here so that it can be merged. Not marking it as a draft PR because i'd still say this is ready for review.

Request issue: https://github.com/JordanMartinez/purescript-cookbook/issues/232


milesfrain:
Looks good.
Made a small change to just point to the original recipe (thanks for the citation though). Make sure to re-pull your branch to pickup this latest addition.

JordanMartinez:
Yeah, I think this looks good, too.

packages.dhall
```diff
@@ -3,6 +3,71 @@ let upstream =
 
 let overrides = {=}
 
-let additions = {=}
```
milesfrain, line 6: Obvious to all of us, but just noting for the record that we'll merge once we can eliminate this change.

JordanMartinez: I think we should add a tag to PRs that are basically "good to merge" once their dependencies are included in the package set. Thoughts?

milesfrain: Sounds good. Added the `package-set` label. Feel free to edit that label if you had something else in mind.

3ddyy:
PR to package-sets submitted (https://github.com/purescript/package-sets/pull/678), will just have to wait for that to be merged, and then i can get this one ready.

JordanMartinez:
Looks like the package got merged and a new release including it was made. Can you update the package set to the latest one and push the changes here? That should get the build working and then we can merge it.

3ddyy:
That should be it!

PR #242, opened 2020-09-24
Add FileUploadReactHooks

ptrfrncsmrph:
Partially addresses #157


recipes/FileUploadReactHooks/src/Main.purs
```diff
@@ -0,0 +1,50 @@
...
+      render (fileUploadComponent {}) (toElement b)
+
+mkFileUploadComponent :: Component {}
+mkFileUploadComponent = do
+  component "FileUploadComponent" \_ -> React.do
+    fileList /\ setFileList <- useState' []
+    let
+      handleChange t =
+        join <$> traverse HTMLInputElement.files (HTMLInputElement.fromEventTarget t)
+          >>= traverse_ (FileList.items >>> map File.name >>> setFileList)
```
ptrfrncsmrph, line 39: I think this could be improved, at least readability-wise 🤔

JordanMartinez: Yeah...

Why not move the `join` into its own block followed by a `do`?
```
handleChange t = join do
  -- code
```
If `traverse` was replaced with `for`, would that make it easier to read?

ptrfrncsmrph: > If `traverse` was replaced with `for`, would that make it easier to read?

Possibly? But I don't know if there's any way of removing the `join` (aside from explicit pattern matching), since
```haskell
traverse files (fromEventTarget t)
```
gives me an `Effect (Maybe (Maybe FileList))`.

I could do this:
```haskell
handleChange t =
  traverse HTMLInputElement.files (HTMLInputElement.fromEventTarget t)
    >>= join
    >>> traverse_ (FileList.items >>> map File.name >>> setFileList)
```
but maybe better to be more explicit
```haskell
      handleChange t = case HTMLInputElement.fromEventTarget t of
        Nothing -> pure unit
        Just fileInput -> do
          maybeFiles <- (HTMLInputElement.files fileInput)
          for_ maybeFiles
            (FileList.items >>> map File.name >>> setFileList)
```
?

I think I like the latter, I feel like if I wrote it like I had it originally I would have trouble understanding it if I came back to it like a week from now 😅.

recipes/FileUploadReactHooks/src/Main.purs
```diff
@@ -34,9 +33,12 @@ mkFileUploadComponent = do
...
     let
-      handleChange t =
-        join <$> traverse HTMLInputElement.files (HTMLInputElement.fromEventTarget t)
-          >>= traverse_ (FileList.items >>> map File.name >>> setFileList)
+      handleChange t = case HTMLInputElement.fromEventTarget t of
+        Nothing -> pure unit
+        Just fileInput -> do
+          maybeFiles <- (HTMLInputElement.files fileInput)
+          for_ maybeFiles
+            (FileList.items >>> map File.name >>> setFileList)
```
JordanMartinez, line 41: Does this compile? If so, it gets rid of that case statement.
```purescript
handleChange t =
  for_ (HtmlInputElement.fromEventTarget t) \fileInput -> do
    maybeFiles <- (HTMLInputElement.files fileInput)
    for_ maybeFiles (setFileList <<< map File.name <<< FileList.items)
```

ptrfrncsmrph: 🎉 It does, and I like it 👍 678f287502f603c3e1e4318226bac9745f5c725a

JordanMartinez:
I'm guessing this is ready to merge once CI passes, correct?

ptrfrncsmrph:
> I'm guessing this is ready to merge once CI passes, correct?

Yes, I've tested it & I'm happy with the code 👍

JordanMartinez:
Thanks!

PR #244, opened 2020-12-01
Add SimpleASTParserLog

### LINKED ISSUE #240: simple - but not trivially simple - parser example
**SimplishParser**

```
simpleASTParser
```

**Recipe Description**

I spent my free time this morning working on porting a pretty simple Parsec-based parser from Haskell and, despite knowing Haskell and PureScript quite well i got tripped up on some of the differences and i think a complete beginner would have been tripped up a little more often (by String !== [Char] just to start with). 

Recursive parsers behave differently - i would say a little less elegantly - in PureScript and there are some good notes on this: [here](https://github.com/Thimoteus/SandScript/wiki/2.-Parsing-recursively).

Altho you might think that Haskell would still be the first choice for writing a parser, i think it would absolutely be worthwhile having a recipe here that shows a super simple AST and a parser for it so that people could "copy and extend" without having to wrassle with Haskell translation.

JordanMartinez:
Fixes #240


recipes/SimpleASTParserLog/README.md
```diff
@@ -0,0 +1,15 @@
+# SimpleASTParserLog
+
+This recipe shows how to run a simple "Hello world!" program in either the node.js or web browser console.
```
milesfrain, line 3: Looks like remaining todos are to just update this file and run the readme script.

JordanMartinez: Thanks for the reminder. I pushed those changes just now.

PR #252, opened 2020-12-07
Simplify ComponentsHalogenHooks

milesfrain:
I was initially confused about how `ButtonMessage` was being used, and then realized that the component's output value is simply ignored by the parent when incrementing toggle count. So this can be replaced with `Unit`. Don't know whether this makes things more or less confusing. We could alternatively go with an `Output` alias and some more comments.


JordanMartinez:
Nah, that's fine. Using `unit` is better.

PR #254, opened 2020-12-08
Add TicTacToeReactHooks

ptrfrncsmrph:


recipes/TicTacToeReactHooks/web/index.css
```diff
@@ -0,0 +1,51 @@
...
+}
+
+.game {
+  display: flex;
+  flex-direction: row;
+}
+
+.game-info {
+  margin-left: 20px;
+}
```
ptrfrncsmrph, line 51: I think I'll have to look into translating these styles into the `.purs` file so that it works in TryPureScript

recipes/TicTacToeReactHooks/src/Main.purs
```diff
@@ -0,0 +1,277 @@
...
+              , children: [ 0, 1, 2 ] <#> renderSquare 2
+              }
+          ]
+
+data Action
+  = JumpToStep Int
+  | FillSquare Int Int
+
+type BoardState
+  = Int -> Int -> Square
```
ptrfrncsmrph, line 98: I'm not sure how much of this is worth commenting on, but this is a notable departure from the original example in the React docs, where board state is an array of `'X' | 'O' | null`. Instead of initializing an `Array (Maybe Player)` full of `Nothing` and then having to handle the nested `Maybe (Maybe Player)` when indexing into that array, I thought this representation was a bit easier to work with in PureScript.

milesfrain:
Looks great!

Bummer that the TPS package set is still too old for this example. Hopefully this will be fixed along with 0.14 updates.
https://github.com/purescript/trypurescript/blob/master/staging/packages.dhall
Will create a fixme issue after this recipe is merged.

milesfrain (approved):

JordanMartinez:
Miles approved this, so I'm going to trust him and merge this. Thanks for the example. It looks really nice!

PR #257, opened 2020-12-29
add Capability Pattern recipe

afcondon:
this is minimally changed from my repo version, very much open to changes / comments on this

for example i think that it might be better to name the files after their role (Core, Domain, API-Infrastructure, Main)


JordanMartinez:
I think to make this clearer, the file names should be:
- LayerFour.purs -> Types.purs
- LayerThree.purs -> Application.purs
- Production.purs -> ProductionSync.purs
- ProductionA.purs -> ProductionAsync.purs

afcondon:
That sounds good.

recipes/CapabilityPatternNode/src/Main.purs
```diff
@@ -0,0 +1,43 @@
...
+mainSync :: Sync.Environment -> Effect Unit
+mainSync env = do
+  result <- Sync.runApp program env
+  pure unit
+
+mainTest :: Test.Environment -> Effect Unit
+mainTest env = do
+  assert $ (Test.runApp program env) == "succeeds"
+  log "first test succeeded, now a failing test which will crash"
+  assert $ (Test.runApp program env) == "failing test"
```
milesfrain, line 38: It looks like the failing case is really just testing the `assert` function and not `runApp`.
Is the plan to use different input for the second ("failing") test? In that case, we could probably still use an affirmative assert that expects the returned error value. Then we'd avoid crashing in CI.
We could alternatively just stick with the single passing test case.

recipes/CapabilityPatternNode/README.md
```diff
@@ -0,0 +1,15 @@
+# CapabilityPatternNode
+
+A skeletal version of an application structuring pattern
+
+See the readme in https://github.com/afcondon/purescript-capability-pattern
```
JordanMartinez, line 5: I'm wondering whether we should inline that repo's readme content here. @milesfrain your thoughts?

recipes/CapabilityPatternNode/src/Main.purs
JordanMartinez: We could also rename `nodeSupported.md` to `nodeSupportedSkipCI.md` and that will no longer run this program during CI. I think there is value showing both successful and failing tests.

recipes/CapabilityPatternNode/README.md
milesfrain: Yes. That would probably be easier to maintain

recipes/CapabilityPatternNode/src/Main.purs
afcondon: the assert could be clearer - you run the same program but expect a different result and you get a failure

i think a recipe of this size and complexity is right at the edge of what can reasonably be in a cookbook but at the same time, AFAICT, something like this really would be the minimal case for a real world application structure so i think it does belong (unless and until there's like Cookbook 2: Advanced PureScript Recipes)

for that reason it does seem as if the failing test belongs

recipes/CapabilityPatternNode/README.md
afcondon: i'll do that now, so

afcondon: I'll edit it to make it less "personal repo chatty" and more "cookbook semi-formal"

recipes/CapabilityPatternNode/src/Main.purs
milesfrain: > I think there is value showing both successful and failing tests.

Yes, but the current form doesn't add any value. It's pretty much the same as:
```purs
assert $ "success" == "failing test"
```

We should instead do something like:
```purs
 assert $ Test.runApp bad_program bad_env == "~expected output for bad input"
```
Where the assert is still happy and does not crash.

afcondon: so `bad_program bad_env` just runs and produces a string  "~expected output for bad input"?
or is the tilde significant in some way?  it may be that i'm ignorant of how test is supposed to work somehow but i'm not seeing how that shows us running the same program in all situations (production, production aff, test).

milesfrain: Going for ["approximate"](https://en.wikipedia.org/wiki/Tilde#Common_use) with that tilde. The actual error should be more precise, so could be an `Either`, or another ADT.

I don't know what should to do for testing this outside of Test, or if that's even what should be done. I mostly just wanted to point out that tests should be designed to avoid `assert` crashes (even when checking for expected failures).

milesfrain: Just noticed this message:
> you run the same program but expect a different result and you get a failure

That sounds like a test for purity. I don't know if that's exactly what we need to be testing for, but if we do wish to test for that, it would be more correct to just re-run the initial test with the same expected output. That has two benefits:
- Assert won't crash. CI still works. We can continue to test for more things.
- It's a stronger check. For example, if `runApp` was not pure, it would be indistinguishable from the pure version with that test in almost all (infinite - 1) possible outcomes. The only distinguishable case is when `runApp` happens to return "failing test", but you'd still get the identical behavior (as far as `assert` is concerned) to the pure version by impurely returning anything else, e.g. "whatever25235".

JordanMartinez: How about we just comment out the last two lines? Someone who clones the repo and tries it out themselves can verify that it fails.

afcondon: That sounds the easiest path - obviously we can revisit it and make it better in the future, too.

JordanMartinez:
Thanks @afcondon!

PR #259, opened 2020-12-31
add an affjax POST recipe

### LINKED ISSUE #247: Post requests with PureScript Affjax and Argonaut
**Recipe Name**

```
AffjaxPostExample
```

**Recipe Description**

Update of blogpost here: https://codingstruggles.com/purescript/post-request-with-purescript-affjax-and-argonaut.html

The post is pretty self-documenting but, being 4 years old, is written using Eff { row } instead of Effect, seems like a very easy port. I'll put a comment on the blog letting the author know i've created this issue.

> In this post I will explain how make a simple ajax call in PureScript, using Affjax and Argonaut.

>For this example I’m going to use http://jsonplaceholder.typicode.com/ as my API endpoints. We will make a POST call to the /posts endpoint to create a new post.

afcondon:
Fixes #247


recipes/AffjaxPostNode/src/Main.purs
```diff
@@ -0,0 +1,51 @@
...
+  case result of
+
+    Left err       -> log $ "POST " <> endpoint <> " response failed to decode: " 
+      <> AX.printError err
+
+    Right response -> log $ "POST " <> endpoint <> " response: "
+      <> J.stringify response.body
+
+
+-- expected response: (at least iff `endpoint` is available to you)
```
JordanMartinez, line 50: typo `iff`

recipes/AffjaxPostNode/README.md
```diff
@@ -0,0 +1,21 @@
+# AffjaxPostNode
+
+Performs a simple HTTP Post request using Affjax library.
+
+The post data that is sent with the request makes use of argonaut to reduce boilerplate
```
JordanMartinez, line 5: Two comments here. Could you provide a link to `argonaut`? Otherwise, the user might not know that that's the name of a library.

Second, I'm wondering whether we should clarify that `reducing boilerplate` via `argonaut` comes with a tradeoff that Gary describes in his blog post here: https://code.slipthrough.net/2018/03/13/thoughts-on-typeclass-codecs/

recipes/AffjaxPostNode/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Expected Behavior:
+
+Prints to the console:
+
+`-- POST http://jsonplaceholder.typicode.com/posts response: {"userId":22,"title":"title","id":101,"body":"body"}`
+
+## Dependencies used
+
+[xhr2](https://www.npmjs.com/package/xhr2)
```
JordanMartinez, line 17: Can you clarify that this is only needed if using `affjax` in a Node environment?

JordanMartinez (changes requested):
Looks good. Thanks! Just a few minor points that need to be addressed.

recipes/AffjaxPostNode/src/Main.purs
afcondon: i meant it as "iff", ie "if and only if" but if that shorthand is cryptic - in fact, even if it isn't i'll change it.

recipes/AffjaxPostNode/README.md
afcondon: should i be committing the changes to `package.json` and `package-lock.json` for the `xhr2` dependency, btw?

i may have missed it but didn't see in the contribute file whether that's desired or not...

milesfrain: You'll just need to add `xhr2` to [`package.json`](https://github.com/JordanMartinez/purescript-cookbook/blob/master/package.json). We're not tracking `package-lock.json`.
For example, you can see the `big-integer` dependency from the [BigIntJs recipe](https://github.com/JordanMartinez/purescript-cookbook/tree/master/recipes/BigIntJs) listed in that file.
Feel free to add some additional clarifications to the [contributing docs section](https://github.com/JordanMartinez/purescript-cookbook/blob/master/CONTRIBUTING.md#all-recipes-that-use-the-same-npm-dependency-must-use-the-same-version-across-all-recipes) on this step.

afcondon: have added this, will PR the CONTRIBUTING.md in separate PR

recipes/AffjaxPostNode/src/Main.purs
```diff
@@ -0,0 +1,51 @@
...
+    postdata :: Post
+    postdata =  { id    : Nothing -- note how this returns filled-in if service works
+                , title : "title"
+                , body  : "body"
+                , userId: 22 }
+
+    endpoint :: String
+    endpoint = "http://jsonplaceholder.typicode.com/posts"
+    
+  result <- AX.post ResponseFormat.json endpoint (Just $ json $ encodeJson postdata)
```
JordanMartinez, line 40: Missed this in my first review, but why are you using `encodeJson postData` rather than `postToJson`?

recipes/AffjaxPostNode/src/Main.purs
```diff
@@ -0,0 +1,51 @@
...
+  case result of
+
+    Left err       -> log $ "POST " <> endpoint <> " response failed to decode: " 
+      <> AX.printError err
+
+    Right response -> log $ "POST " <> endpoint <> " response: "
+      <> J.stringify response.body
```
JordanMartinez, line 47: Similarly here. Perhaps this should be:
```purescript
Right response -> do
  log $ "POST " <> endpoint <> " response: " <> J.stringify response.body
  log $ ">>> " <> show (jsonToPost response.body)
```

JordanMartinez (changes requested):
I missed these in my initial review, but shouldn't you be using `postToJson` and `jsonToPost` in this recipe?

recipes/AffjaxPostNode/src/Main.purs
afcondon: oh, good catch, i was actually going back and forth on including the synonyms or the type qualification or the explicit written `encode` and `decode`, i'll fix it now

recipes/AffjaxPostNode/src/Main.purs
afcondon: d'you mean:
```haskell
    Right response -> do
      log $ "POST " <> endpoint <> " response: " <> show (postFromJson response.body)
```
??

we don't need to show the `J.stringify` i think?

afcondon:
D'you think it would be worth putting the explicit encode/decode instances back in, in comments?

recipes/AffjaxPostNode/src/Main.purs
JordanMartinez: No, I really did mean what I wrote above. The first line will show the JSON in a string format. The second line will convert the body into our data type and then use that type's `Show` instance to convert it into a String. 

By having both, it shows the JSON you got back and the resulting data type you produced by parsing the JSON via `decodeJson`

recipes/AffjaxPostNode/src/Main.purs
JordanMartinez: I think the `postToJson` way is better because it shows the type signatures.

JordanMartinez:
> D'you think it would be worth putting the explicit encode/decode instances back in, in comments?

Not sure I follow. Could you clarify?

afcondon:
i can include this so that recipe shows both approaches

```haskell
-- Version 2 - alternative implementation with explicit encodeJson and decodeJson follows
-- this approach tends to protect against ugly versioning problems
-- see https://code.slipthrough.net/2018/03/13/thoughts-on-typeclass-codecs/

data Post2 = Post2 Post
instance showPost2 :: Show Post2 where
  show (Post2 post) = show post

instance encodeJsonPost :: EncodeJson Post2 where
  encodeJson (Post2 post)
    = "id"      := post.id
    ~> "title"  := post.title
    ~> "body"   := post.body
    ~> "userId" := post.userId
    ~> jsonEmptyObject


instance decodeJsonPost :: DecodeJson Post2 where
  decodeJson json = do
    obj    <- decodeJson json
    id     <- obj .:? "id"
    title  <- obj .: "title"
    body   <- obj .: "body"
    userId <- obj .: "userId"
    pure $ Post2 { id, title, body, userId }


version2 :: Effect Unit
version2 = launchAff_ $ do
  let 
    postdata :: Post
    postdata =  { id    : Nothing -- note how this returns filled-in if service works
                , title : "title"
                , body  : "body"
                , userId: 22 }

    endpoint :: String
    endpoint = "http://jsonplaceholder.typicode.com/posts"
    
  result <- AX.post ResponseFormat.json endpoint (Just $ json $ encodeJson $ Post2 postdata)
  case result of

    Left err -> log $ "POST " <> endpoint <> " response failed to decode: " <> AX.printError err

    Right response ->
      log $ "Second version, explicit instances of codecs "
        <> "POST " <> endpoint <> " response: " <> show ((decodeJson response.body) :: Either JsonDecodeError Post2)
```

afcondon:
i've added the explicit encode / decode - see what you think, i think, on reflection it's actually much more useful that way, better than, say, two recipes. README has been updated to reflect this

Also contains your suggestion to show both JSON and the decoded object.

JordanMartinez:
Ah, now I see what you mean. I don't think the second version should be included in here. Rather, I think readers would be better served by looking at an `argonaut` version of [ValueBasedJsonCodecLog](https://github.com/JordanMartinez/purescript-cookbook/tree/master/recipes/ValueBasedJsonCodecLog) because that would highlight the pros/cons of each approach.

The purpose of this recipe is to show how to use Affjax to do a POST request. I think `argonaut` being used here makes sense because it removes the noise of the codec boilerplate.

Also, you shouldn't be using `data Post2 = Post2 Post` as this leads to unnecessary boxing. Rather, you should use a newtype (i.e. `newtype Post2 = Post2 Post`).

afcondon:
Okay, that’s perfect I’ll take it back to simpler version this morning. Good point on newtype too, thanks

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

afcondon (commented):
thanks for spotting those

JordanMartinez (approved):

JordanMartinez:
Thanks @afcondon!

PR #264, opened 2021-01-18
Bump `easy-purescript-nix`

ptrfrncsmrph:
`packages.dhall` specifies PureScript 0.13.8, while the `easy-purescript-nix` commit hash points to 0.13.6, so trying to build the project in `nix-shell` fails.

### Before
Screen Shot 2021-01-18 at 1 36 19 AM

### After
Screen Shot 2021-01-18 at 1 37 39 AM

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -5,8 +5,12 @@ let
     pkgs.fetchFromGitHub {
       owner = "justinwoo";
       repo = "easy-purescript-nix";
-      rev = "01ae1bc844a4eed1af7dfbbb202fdd297e3441b9";
-      sha256 = "0jx4xb202j43c504gzh27rp9f2571ywdw39dqp6qs76294zwlxkh";
+      # Latest `rev` and `sha256` can be obtained by using
+      # ```
+      # $ nix-prefetch-git https://github.com/justinwoo/easy-purescript-nix.git
+      # ```
+      rev = "c8c32741bc09e2ac0a94d5140cf51fa5de809e24";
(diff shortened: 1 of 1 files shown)


milesfrain (approved):
I haven't jumped on the nix train yet, but I trust this is a change for the better.
Also, those are some nice looking shadowed screenshots. Was wondering if that was a new GH feature, but guessing you just grabbed a bit of desktop too.

ptrfrncsmrph:
> I haven't jumped on the nix train yet, but I trust this is a change for the better.

Yeah, I don't know, it seems like keeping the `shell.nix` up to date might be a bit of a burden if the core maintainers of this project aren't using Nix, so alternatively could just scrap the Nix stuff 🤷 . For Nix users, `nix-prefetch-git https://github.com/justinwoo/easy-purescript-nix.git` will get the latest `rev` and `sha256`, so this at least adds those instructions if it gets out of sync in the future. I wonder if it's something that could be automated in CI.

> Also, those are some nice looking shadowed screenshots. Was wondering if that was a new GH feature, but guessing you just grabbed a bit of desktop too.

I'm on a Mac, that's just the default screenshot look when you capture a window. It is nice.

JordanMartinez:
Yeah, `nix` is a great idea but something I just don't use right now. If we did support it, I would prefer to add a disclaimer of "use nix to build this project at your own risk; maintainers have not vetted what the nix code does" or something to that effect. If we don't do that, then dropping it entirely would be best.

JordanMartinez:
I'm not sure I want to keep `nix` in the future, but I think this is safe to merge.

PR #271, opened 2021-02-07
Add RoutingHashReactHooks recipe

ptrfrncsmrph:
Similar to the `RoutingHashHalogenClassic` recipe, but I put everything in one module to make it compatible with TryPureScript. I can split things up if that makes more sense.


milesfrain (approved):

JordanMartinez:
Thanks all!

ptrfrncsmrph:
Oops! It seems that this is _not_ compatible with TryPureScript (though I copied the implementation from an existing [gist of mine](https://try.purescript.org/?gist=1c5c2f222f723acefae637df3f6df721) that _does_ seem to work with TryPureScript). Not sure what the issue is there, but I'll make a patch to add the `tryUnsupported.md` file as a flag.

PR #272, opened 2021-02-08
Add RoutingPushReactHooks

ptrfrncsmrph:
Very much the same UI as `RoutingHashReactHooks`, but using `Routing.PushState`


recipes/RoutingPushReactHooks/src/Main.purs
```diff
@@ -0,0 +1,202 @@
...
+-- | provided (signalling that this is not a use case we want to support). We've
+-- | done similar, by wrapping our context value in `Maybe` and using `Nothing`
+-- | as the case that we pattern-match on to trigger the error.
+routerContext :: ReactContext (Maybe RouterContextValue)
+routerContext =
+  -- | We are using `unsafePerformEffect` so that we can have a "global"
+  -- | instance of this context that we can use directly in our
+  -- | `useRouterContext` hook.
+  Effect.Unsafe.unsafePerformEffect do
+    React.createContext Nothing
```
ptrfrncsmrph, line 165: Now I'm not sure if any of this is the recommended approach, after reading through this: https://github.com/spicydonuts/purescript-react-basic-hooks/issues/41 😅 Will have to think on how best to refactor this.

ptrfrncsmrph: I've refactored to avoid `unsafePerformEffect`

JordanMartinez:
Thanks!

PR #273, opened 2021-02-08
RoutingHashReactHooks does not work in TryPureScript

ptrfrncsmrph:
This recipe causes odd behavior in TryPureScript (as first noted here: https://github.com/JordanMartinez/purescript-cookbook/pull/271#issuecomment-774830288, and pictured below) where clicking on a link will re-direct such that the `<iframe>` renders a nested instance of TryPureScript. I've added a `tryUnsupported.md` and removed the **Try** link from the `README` table of contents. Interestingly this exact code will work if run as a [gist in TryPureScript](https://try.purescript.org/?gist=80c9852e55cd5b709d6c53822bc734c1). This seems to be unexpected behavior; should I open an actual issue to track this?


JordanMartinez:
I want to say this is an issue with Try PureScript, but I'm not really sure what is causing it.

Either way, I think we should merge this to prevent this unexpected behavior.

milesfrain:
try.ps.ai generates the iframe a bit differently than the original site, and this is definitely a regression. Thanks for reporting. I'll dig into this whenever I get around to making another pass at merging with upstream.

PR #280, opened 2021-07-15
Bump PureScript to 0.14.3

ptrfrncsmrph:
All told, only ~17~ 14 of 80 ended up in the `broken` pile which doesn't seem too bad.

---

- [x] AceEditorHalogenHooks
  :x: Move to `broken`
  ```
  Module Halogen.Query.EventSource was not found.
  ```
- [x] Add RemoveEventListenerJs
  :pencil2: Add missing dependencies to `spago.dhall`
- [x] AffGameSnakeJs
  :x: Move to `broken`
  ```
  [error] The following packages do not exist in your package set:
  [error]   - canvas-action
  [error]   - game
  ```
- [x] AffjaxPostNode
  :pencil2: Add missing dependencies to `spago.dhall`
- [x] BasicHalogenHooks
  :pencil2: Remove `HH.HTML` as parameter to `Component`
  :pencil2: Update event handler signature from `MouseEvent -> Maybe i` to `MouseEvent -> i`
- [x] BehaviorSuperCircleJs
  :x: Move to `broken`
  ```
  [error] The following packages do not exist in your package set:
  [error]   - behaviors
  ```
- [x] BigIntJs
  :white_check_mark: Already works
- [x] BookHalogenHooks
  :pencil2: Remove `HH.HTML` as parameter to `Component`
- [x] BookReactHooks
  :white_check_mark: Already works
- [x] ButtonsHalogenHooks
  :pencil2: Remove `HH.HTML` as parameter to `Component`
  :pencil2: Update event handler signature from `MouseEvent -> Maybe i` to `MouseEvent -> i`
- [x] ButtonsReactHooks
  :white_check_mark: Already works
- [x] CapabilityPatternNode
  :white_check_mark: Already works
- [x] CardsHalogenHooks
  :pencil2: Remove `HH.HTML` as parameter to `Component`
  :pencil2: Update event handler signature from `MouseEvent -> Maybe i` to `MouseEvent -> i`
  :pencil2: Change `NonEmpty Array` to `NonEmptyArray`
- [x] CardsReactHooks
  :pencil2: Change `NonEmpty Array` to `NonEmptyArray`
- [x] CatGifsHalogenHooks
  :pencil2: Remove `HH.HTML` as parameter to `Component`
  :pencil2: Update event handler signature from `MouseEvent -> Maybe i` to `MouseEvent -> i`
- [x] CatGifsReactHooks
  :white_check_mark: Already works
- [x] ClockReactHooks
  :white_check_mark: Already works
[...]

PR #283, opened 2021-07-18
Restore broken Components*HalogenHooks recipes

ptrfrncsmrph:
In #280 I had moved the three `Components*HalogenHooks` recipes to `broken`, this just updates them to Halogen 6 and restores them to `recipes`.


JordanMartinez (approved):
Thanks again!

PR #288, opened 2022-02-14
chore: giphy cat api has updated, updating CatGifsReactHook to match

neilhan:
CatGifsReactHook was trying to extract a gif url that does not match the latest giphy respones. 
Updated the getRandomCatGif function to work with latest giphy API


ptrfrncsmrph (approved):
Looks good to me 👍 thanks!

Looks like a similar fix is needed for the counterpart Halogen recipe, I'll open an issue for that.

JordanMartinez:
Thanks!

Issue #291, opened 2022-06-18
Update cookbook recipes to PureScript 0.15.2


JordanMartinez:
At some point, this cookbook should be updated to `0.15.2` now that we can use ES modules.

Besides updating the recipes, the current blocker is how to run the examples in a web browser. Currently, it's done via `try.ps.ai`. However, that was setup by Miles and I don't want to maintain it even if I was granted access / shown how to maintain it. It would be easier to just use `try.purescript.org` directly, but the examples can't be loaded due to https://github.com/purescript/trypurescript/issues/174

So, I think dropping the `Main` module requirement in Try PureScript would need to be done first and then we could continue with this.

JordanMartinez:
Closed by #292

PR #292, opened 2022-07-12
Update repository to PureScript 0.15.4

andys8:
## Changes

* Fix code where necessary due to breaking changes in libraries (colors were deleted, modules renamed, affjax change)
* Change index.js and index.html to work with parcel 2 and es modules
* Removed all build warnings

## Broken

- `HelloConcur`: No recent activity for `concur` in 2022
- `ReadPrintFileContentsConcur`: No recent activity for `concur` in 2022
- `PayloadHttpApiNode`: There is a branch open to update https://github.com/hoodunit/purescript-payload/pull/27

## Related

Solves https://github.com/JordanMartinez/purescript-cookbook/issues/291


package.json
```diff
@@ -1,8 +1,9 @@
 {
   "license": "MIT",
   "devDependencies": {
-    "parcel-bundler": "^1.12.5",
-    "purescript": "^0.14.3",
+    "parcel": "^2.6.0",
+    "process": "^0.11.10",
```
JordanMartinez, line 5: Could you clarify why `process` is being added here?

recipes/AddRemoveEventListenerJs/src/Main.purs
```diff
@@ -85,7 +85,7 @@ main = do
       -- In this approach, we use a better approach to add an event listener
       -- where we don't need to store a reference to the listener since
       -- it is still in scope.
-      removeListener <- addBetterListener METypes.click false buttonMain \e -> do
+      removeListener <- addBetterListener METypes.click false buttonMain \_e -> do
```
JordanMartinez, line 88: Shouldn't this just be `\_ ->` if the `e` isn't used?

recipes/SimpleASTParserLog/src/Main.purs
```diff
@@ -89,7 +89,7 @@ parseExpr = fix \parseInfix -> do
       leftOp <- try $ between skipSpaces skipSpaces parseBinaryOperator
       nextPart <- parseInfix
       pure case nextPart of
-        UnaryOp right -> do
+        UnaryOp _right -> do
```
JordanMartinez, line 92:
```diff
- UnaryOp right -> do
+ UnaryOp _ -> do
```

JordanMartinez (changes requested):

JordanMartinez:
Should we continue using `parcel`? `esbuild` can also be used to build, bundle, and serve code. Since it's much faster, it would improve our CI times as well.

recipes/AddRemoveEventListenerJs/src/Main.purs
andys8: Unused, named parameters as feature was intentionally used here, to give a bit of context what's ignored (e.g. `event`, `rec` and so on). I can drop it and use `_` instead.

package.json
andys8: Good catch, since I can't recall I must have added it by accident.

andys8:
> Should we continue using `parcel`? `esbuild` can also be used to build, bundle, and serve code. Since it's much faster, it would improve our CI times as well.

I assume it could work. We're essentially only using parcel arguments that look like they can be replaced by `esbuild`.

```shell
parcel $(call webHtml,$*) --dist-dir $(call webDistDir,$*) --open
parcel build $(call webHtml,$*) --dist-dir $(call webDistDir,$*) --no-optimize --no-source-maps
parcel build $(call prodHtml,$*) --dist-dir $(call prodDistDir,$*)
```

Priority wise (compared to React v18 and try.purescript.org) and because of the amount of changes I wouldn't do/try this change with this pull request, though.

recipes/SimpleASTParserLog/src/Main.purs
andys8: a1782dcd96e7d41dcce1b99459184976dbbb93ae

recipes/LifecycleHalogenHooks/src/Main.purs
```diff
@@ -28,7 +28,7 @@ rootComponent
   :: forall unusedInput unusedQuery unusedOutput anyMonad
    . MonadEffect anyMonad
   => H.Component unusedQuery unusedInput unusedOutput anyMonad
-rootComponent = Hooks.component \rec _ -> Hooks.do
+rootComponent = Hooks.component \_rec _ -> Hooks.do
```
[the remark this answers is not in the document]
andys8: a1782dcd96e7d41dcce1b99459184976dbbb93ae

package.json
andys8: Okay, I think I found it.

It will be added automatically as part running `make testAllCI`. `SignalRenderJs` is using process. parcel figures out and magically auto installs it.


image

https://parceljs.org/features/node-emulation/#polyfilling-%26-excluding-builtin-node-modules

image

Should we add it back in or let parcel add it (creating a diff) for the user in this case?

JordanMartinez: Ok, thanks for clarifying. Let's just keep `process` as a dependency then.

andreas-ebot7:
Added `process` back in. There was an issue on CI, too, because of this dependency missing. Btw. I can't trigger CI runs myself, so I hope this one will go though.

image

andys8:
Added `process` back in. There was an issue on CI, too, because of this dependency missing. Btw. I can't trigger CI runs myself, so I hope this one will go though.

image

package.json
andys8: Added back in

JordanMartinez (commented):
Thanks! Looks like we just need to update Try PureScript, and then this will be ready to go.

JordanMartinez:
Try PureScript has been updated:
- module names can be something other than `Main` so long as it doesn't clash with anything else in the package set
- a `div id="root'` element was added to the `frame.html`, so that one can hook into that element.

package.json
```diff
@@ -1,8 +1,10 @@
 {
   "license": "MIT",
   "devDependencies": {
-    "parcel-bundler": "^1.12.5",
-    "purescript": "^0.14.3",
+    "esbuild": "^0.14.49",
+    "parcel": "^2.6.0",
+    "process": "^0.11.10",
+    "purescript": "^0.15.3",
     "spago": "^0.20.3"
```
JordanMartinez, line 8: These dependencies need to be updated:
- `spago`: 0.20.9 - running any of the recipes will fail because this version of Spago runs code as thought it were CJS modules.
- `purescript`: we should bump this to `0.15.4` now.

JordanMartinez (changes requested):
Besides the changes described below, the `generateRecipeTable.sh` file **would** need to update line 31 to:
```diff
-      tryLink="[try](https://try.ps.ai/?github=JordanMartinez/purescript-cookbook/master/recipes/$base/src/Main.purs)"
+      tryLink="[try](https://try.purescript.org/?github=/JordanMartinez/purescript-cookbook/master/recipes/$base/src/Main.purs)"
```

However, I was testing out the "try" links (using `andys8` in the URL above) and found an issue with Try PureScript when `big-integer` was being used. Looks like there's another TryPureScript-related issue that needs to be fixed.

So as not to delay this PR any further, how about the script just remove the `try` link? I'll see what needs to be done to fix TPS and add that link back in later.

package.json
andys8: `^` is a version range and no `package-lock.json` is checked in. Therefore any of the listed updates are included or will be installed.

If specific version should be used, I would rather switch to fixed versions and/or checking in the `package-lock.json`.

package.json
```diff
@@ -1,8 +1,10 @@
...
   },
   "dependencies": {
```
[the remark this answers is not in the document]
andys8: https://github.com/JordanMartinez/purescript-cookbook/pull/292#discussion_r922500876

packages.dhall
```diff
@@ -1,5 +1,6 @@
 let upstream =
-      https://github.com/purescript/package-sets/releases/download/psc-0.14.3-20210716/packages.dhall sha256:1f9af624ddfd5352455b7ac6df714f950d499e7e3c6504f62ff467eebd11042c
+      https://github.com/purescript/package-sets/releases/download/psc-0.15.2-20220706/packages.dhall
```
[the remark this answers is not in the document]
andys8: Sure thing 6519649c2ab4577806cea1ab6e4032b7e22fdfce

package.json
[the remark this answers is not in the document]
andys8: Btw. I can offer to prepare a pull request for react v18. Since the changes are already merged (https://github.com/lumihq/purescript-react-basic-dom/pull/44) the repository can be prepared by overriding the dependency, updating the react examples and be merged as soon as the package-set is ready.

package.json
JordanMartinez: > Therefore any of the listed updates are included or will be installed.

You're right. Hmm.... When I ran `make installDeps`, it didn't update my dependencies. When I `rm -rf node_modules && make installDeps` then it updated them.

I don't think we should use fixed versions because then we'd have to update this repo every time a new release was made, so the version ranges make sense.

JordanMartinez (approved):
Thanks for all your work!

package.json
andys8: One explanation could be that you created a `package-lock.json` with a previous installation. The file is git-ignored, but still exists locally, and would be respected. I think npm wouldn't just replace installed versions if they match the `package-lock.json` if there are newer versions available.

JordanMartinez: Ah... Yeah, I think that's what happened.

Issue #198, opened 2020-08-06
Compact and warning-free React recipes


milesfrain:
Follow-up to #197 

The existing code produces a warning in the dev console with [this strategy](https://try.ps.ai/?gist=f42dc4b8e49846de036d0556d8ec4e11):
```purs
main :: Effect Unit
main = do
  body <- body =<< document =<< window
  case body of
    Nothing -> throw "Could not find body."
    Just c -> do
      buttons <- mkButtons
      render (buttons {}) (HTMLElement.toElement c)
```

Here's a [warning-free option](https://try.ps.ai/?gist=f1d001df5581561115ef289015f4d339) that's a little more verbose:
```purs
-- App setup
main :: Effect Unit 
main = do
  root <- createRootElement
  buttons <- mkButtons
  render (buttons {}) root

-- DOM helper, creates an element we can mount
-- the React app into
createRootElement :: Effect Element
createRootElement = do
  doc <- document =<< window
  body <- maybe (throw "Could not find body element") pure =<< HTMLDocument.body doc
  root <- createElement "div" $ HTMLDocument.toDocument doc
  root # Element.setClassName "app-root"
  void $ appendChild (Element.toNode root) (HTMLElement.toNode body)
  pure root
```

It would be most ideal to have a utility function that handles this common pattern:
``` purs
main :: Effect Unit 
main = renderInRoot $ mkButtons {}
```

Tracking a proposal to add this utility function to the React library in https://github.com/lumihq/purescript-react-basic-dom/issues/8

Also wondering if we should make code that's reused between recipes more generic (even if `scripts/newRecipe.sh`) handles this renaming for us. For example, removing "buttons" where possible:
```purs
myComponent :: Component {}
myComponent = do
  component "MyComponentName" \_ -> React.do
    -- button-specific code here

main :: Effect Unit 
main = renderInRoot $ myComponent {}
```

JordanMartinez:
> Tracking a proposal to add this utility function to the React library in lumihq/purescript-react-basic-dom#8

I think it would be better to get this utility function merged and then update the cookbook here.

> Also wondering if we should make code that's reused between recipes more generic (even if scripts/newRecipe.sh) handles this renaming for us. For example, removing "buttons" where possible:

[...]

PR #294, opened 2022-07-16
React v18 update

andys8:
This is the suggested update path to react v18:

* Rendering into `#root` (solves Solves https://github.com/JordanMartinez/purescript-cookbook/issues/198)
* Using `createRoot` as recommented in react v18
* Rename component to generic name (`app`), for consistency across examples.

If this change is good to go, I can roll it out to all `ReactHooks` examples.

```
./BookReactHooks/
./ButtonsReactHooks/
./CardsReactHooks/
./CatGifsReactHooks/
./ClockReactHooks/
./ComponentsInputReactHooks/
./ComponentsMultiTypeReactHooks/
./ComponentsReactHooks/
./DragAndDropReactHooks/
./FileUploadReactHooks/
./FormsReactHooks/
./GroceriesReactHooks/
./HelloReactHooks/
./ImagePreviewsReactHooks/
./NumbersReactHooks/
./PositionsReactHooks/
./RoutingHashReactHooks/
./RoutingPushReactHooks/
./ShapesReactHooks/
./TextFieldsReactHooks/
./TicTacToeReactHooks/
./TimeReactHooks/
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,16 +2,16 @@
   "license": "MIT",
   "devDependencies": {
     "esbuild": "^0.14.49",
-    "parcel": "^2.6.0",
+    "parcel": "^2.6.2",
     "process": "^0.11.10",
-    "purescript": "^0.15.3",
-    "spago": "^0.20.3"
+    "purescript": "^0.15.4",
+    "spago": "^0.20.9"
   },
   "dependencies": {
-    "ace-builds": "^1.4.11",
-    "big-integer": "^1.6.48",
-    "react": "^16.14.0",
-    "react-dom": "^16.14.0",
-    "xhr2": "^0.2.0"
+    "ace-builds": "^1.7.1",
+    "big-integer": "^1.6.51",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "xhr2": "^0.2.1"
   }
 }
--- a/packages.dhall
+++ b/packages.dhall
@@ -18,7 +18,7 @@ let overrides =
           , "web-file"
           , "web-html"
           ]
-        , repo = "https://github.com/andys8/purescript-react-basic-dom.git"
+        , repo = "https://github.com/lumihq/purescript-react-basic-dom.git"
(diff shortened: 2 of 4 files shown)


JordanMartinez:
This looks good, but the only issue is the override in the `packages.dhall` file. (See https://github.com/JordanMartinez/purescript-cookbook/blob/master/CONTRIBUTING.md#only-use-packages-in-the-latest-package-set-release)

I get that the changes needed are in `master` for that library, but I'd still prefer to wait until an official release is made.

Other than that, yeah, this seems fine to me.

ptrfrncsmrph (approved):

andys8:
Yes, the override will be dropped and replaced by package set update before merge. Converted to draft PR to make clear this won't be merged in its current state.

JordanMartinez:
> Yes, the override will be dropped and replaced by package set update before merge. Converted to draft PR to make clear this won't be merged in its current state.

Sounds good!

JordanMartinez:
Looks like we only need to update the `react` dependency in Try PureScript, update the package set here to remove the override and use the latest release, and then this is good to merge. Is that right?

andys8:
Aaaaand... update all the other react examples. Currently only one of them is modified. This can be another pullrequest though, since the api is deprecated, but not broken and will still work.

The latest package set doesn't have the react update. With the next package set, we'll have it.

* https://github.com/purescript/package-sets/commit/baec24fec2ed9a649cae3d9ccdfa83eb59180e23
* https://github.com/purescript/package-sets/commit/b0fadc9f670bbf51123d1ceb67cae7431769f207

andys8:
Updated the package set. I would be fine with merging this PR as is:

* Only updating "HelloReactHooks".
* Next PR: Format repository with `purs-tidy`
* Next PR: Update all other react examples.

The current and expected state of the non-updated react examples is to see these warnings. They're exactly what this PR solved, but only for "HelloReactHooks". The examples with warnings still work.

image

image

JordanMartinez:
Thanks!

PR #296, opened 2022-08-10
Format repository with purs-tidy

andys8:
With the goal of consistent formatting and import order across examples the repository is formatted with `purs-tidy`, a dependency and the configuration are added to the repository. Asked upfront if a change like this makes sense and am open to apply changes to `.tidyrc.json` if they're prefered.

## Changes

* Initial repository formatting (`make format`)
* `make formatCheck` command is added and executed on CI

PR #298, opened 2022-08-10
Update 21 react examples to v18

andys8:
Updated react hooks examples to render in `#root` to avoid `body` warning and use new `createRoot` function to avoid deprecation warning.

* Two examples had a different `index.html`
* The component is now consistently `mkApp :: Component {}` in all examples
* Only [`RoutingPushReactHooks`](https://github.com/JordanMartinez/purescript-cookbook/pull/298/commits/aab6f09916d3371cb1f22ff3df700f68f3bbaf09#diff-0f8a853cb9c5828d673267088d4a59779d9fb7d1816a2b1e721a17b53a08e812R43) has a different `main` function because of `routerProvider` with `ReaderT`
* Tiny line break improvements (e.g. one line lets)

## Examples

- [x] HelloReactHooks (updated in previous PR https://github.com/JordanMartinez/purescript-cookbook/pull/294)
- [x] BookReactHooks
- [x] ButtonsReactHooks
- [x] CardsReactHooks
- [x] CatGifsReactHooks
- [x] ClockReactHooks
- [x] ComponentsInputReactHooks
- [x] ComponentsMultiTypeReactHooks
- [x] ComponentsReactHooks
- [x] DragAndDropReactHooks
- [x] FileUploadReactHooks
- [x] FormsReactHooks
- [x] GroceriesReactHooks
- [x] ImagePreviewsReactHooks
- [x] NumbersReactHooks
- [x] PositionsReactHooks
- [x] RoutingHashReactHooks
- [x] RoutingPushReactHooks
- [x] ShapesReactHooks
- [x] TextFieldsReactHooks
- [x] TicTacToeReactHooks
- [x] TimeReactHooks

## Related

* Based on / merge after https://github.com/JordanMartinez/purescript-cookbook/pull/296
* Package set in sync with https://github.com/purescript/trypurescript/pull/294 (`createRoot`)

## Review

Actual diff (without formatting PR) can be better viewed by [comparing this and the base branch](https://github.com/andys8/purescript-cookbook/compare/improvement/format-purs-tidy...andys8:purescript-cookbook:update-react-examples-to-v18)

### DIFF
--- a/packages.dhall
+++ b/packages.dhall
@@ -1,6 +1,6 @@
 let upstream =
-      https://github.com/purescript/package-sets/releases/download/psc-0.15.4-20220723/packages.dhall
-        sha256:efb50561d50d0bebe01f8e5ab21cda51662cca0f5548392bafc3216953a0ed88
+      https://github.com/purescript/package-sets/releases/download/psc-0.15.4-20220808/packages.dhall
+        sha256:60eee64b04ca0013fae3e02a69fc3b176105c6baa2f31865c67cd5f881a412fd
 
 let overrides = {=}
 
--- a/recipes/BookReactHooks/spago.dhall
+++ b/recipes/BookReactHooks/spago.dhall
@@ -10,6 +10,7 @@
   , "react-basic-dom"
   , "react-basic-hooks"
   , "web-html"
+  , "web-dom"
   ]
 , packages = ../../packages.dhall
 , sources = [ "recipes/BookReactHooks/src/**/*.purs" ]
--- a/recipes/BookReactHooks/src/Main.purs
+++ b/recipes/BookReactHooks/src/Main.purs
@@ -9,14 +9,14 @@ import Data.Either (Either(..))
 import Data.Maybe (Maybe(..))
 import Effect (Effect)
 import Effect.Exception (throw)
-import React.Basic.DOM (render)
 import React.Basic.DOM as R
+import React.Basic.DOM.Client (createRoot, renderRoot)
 import React.Basic.Hooks (Component, component)
 import React.Basic.Hooks as React
 import React.Basic.Hooks.Aff (useAff)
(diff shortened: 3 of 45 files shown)


JordanMartinez:
Things left to do here:
- [ ] Fix merge conflicts
- [ ] Fix Try PS issue with react: https://github.com/purescript/trypurescript/pull/294#issuecomment-1213615468

ptrfrncsmrph:
> The component is now consistently `mkApp :: Component {}` in all examples

I started using `Component Unit` in some of the later recipes I added after seeing that's how [Madeline defined `Component`s that take no props](https://github.com/megamaddu/purescript-react-basic-hooks/blob/main/docs/getting-started.md#rendering-your-first-component). I think I had been using `{}` for empty props because I had assumed that props needed to be a record. That said, I don't think there are established idioms here, and I don't know of any reason to prefer `Unit` over `{}` so no need to change it.

ptrfrncsmrph (approved):

JordanMartinez:
> I don't know of any reason to prefer Unit over {} so no need to change it.

Perhaps because `Unit` used to be an empty record, so they were the same thing at runtime?
[Unit's runtime was changed to `undefined`](https://github.com/purescript/purescript-prelude/issues/266) because functions that returned `Unit` (i.e. `{}`) were slower than those that returned `undefined`. We guessed it had something to do with the allocation of an empty record that is immediately thrown away.

andys8:
I think in PureScript one can use anything as props. From a react perspective props are a key value pairs and can be represented as html attributes. Therefore the empty record/object seems a bit more idiomatic than unit, undefined or null to me.

I'd be open to changing if there is a reason to do so, but I think we're good to go as is :)

JordanMartinez (approved):
@andys8 Thank you so much for all your work on both this repo and in the Try PS repo.

PR #304, opened 2023-08-10
Replace giphy API with dog API in `CatGifReactHooks` and `CatGifHalogenHooks` recipes

pete-murphy:
Fixes #297 

`CatGifReactHooks` is now `DogImagesReactHooks`
`CatGifHalogenHooks` is now `DogImagesHalogenHooks`

### Screen recordings

Halogen

https://github.com/JordanMartinez/purescript-cookbook/assets/26548438/c0b94fcc-d4a2-42fe-941b-c7daa707fcc6

React

https://github.com/JordanMartinez/purescript-cookbook/assets/26548438/9f026baf-ed94-4d30-8595-424fce5df7c7

PR #305, opened 2023-08-11
CI: Bump Node version to latest LTS

pete-murphy:
Seeing the following error in the CI build for #304 
```
=== Building HelloLog for the web browser backend ===
/home/runner/work/purescript-cookbook/purescript-cookbook/node_modules/@parcel/utils/lib/index.js:37573
    #getIndex(item) {
             ^

SyntaxError: Unexpected token '('
```
`#getIndex` is using relatively new [private class field](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) syntax. According to MDN, that should be valid in Node v12 so I'm not sure why this is failing with a syntax error. Trying to bump to v18 to see if that helps.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Set up Node toolchain
         uses: actions/setup-node@v1
         with:
-          node-version: "12.x"
+          node-version: "18.x"
 
       - name: Cache NPM dependencies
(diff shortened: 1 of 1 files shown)


pete-murphy:
I'll go ahead and merge this, just a version bump that appears to fix CI following a breaking change in `parcel`

